PR #1, opened 2018-03-16
[UI-9] Prepare icon component

rangoo94:


packages/icon/src/Icon.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cls from 'classnames'
+
+import { prefix } from '@talixo/commons'
+import { icons } from '@talixo/commons/config'
+
+const cName = prefix('icon')
```
michalochman: Do we need `talixo-icon` class name?

packages/icon/src/Icon.js
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @param {object} props
+ * @param {string} [props.name]
+ * @param {string} [props.className]
+ * @returns {React.Element}
+ */
+function Icon (props) {
+  const { className, name, ...passedProps } = props
+
+  const clsName = cls(cName, className, icons.prefix, `${icons.prefix}-${name}`)
```
michalochman: Maybe we could extract `` `${icons.prefix}-${name}` `` to a module aware method similar to `commons.prefix`?

packages/icon/src/StorybookIcons.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  onQueryChange (e) {
+    this.setState({ query: e.target.value })
+  }
+
+  render () {
+    const { query } = this.state
+
+    const groups = Object.keys(icons).map(group => (
```
michalochman: How hard would it be to group material icons into categories like on material design website?

packages/icon/src/Icon.js
rangoo94: We could ignore it - I prepared it to keep prefixes generic, but you're right - it makes output code less readable and bigger.

packages/icon/src/Icon.js
rangoo94: I agree, but we have to think what actually could be in `commons` package to segregate it properly

packages/icon/src/StorybookIcons.js
rangoo94: It is not a problem at all, for performance reasons all icon information is extracted to meta information file, so it could be done easily 👍

packages/icon/src/Icon.js
michalochman: Ok, let's put this to Issues to not forget.

packages/icon/src/Icon.js
michalochman: With `icons.prefix` already added I think it's redundant. Anyway, feel free to kill it or leave it be.

packages/icon/src/StorybookIcons.js
michalochman: I'd go for it, it makes browsing the icons list a little easier, especially for people used to Material Icons.

michalochman:
Whoops, a missclick – sorry.

I have one more question: how can mobile developers use SVG icons from this package?

rangoo94:
In the current state, they can use captured SVG icons available`packages/icon/svg`. I will ask @radzio tomorrow for the preferred way, so we can adjust that.

rangoo94:
Will rewrite it to proper one

PR #3, opened 2018-03-21
Add Icon component

rangoo94:
- Handle custom icons from https://github.com/PublicInMotionGmbH/icon-pack
- Handle Material Design icons


packages/icon/scripts/load-material-icons.js
```diff
@@ -0,0 +1,167 @@
...
+
+  // Gather icons list by codepoints
+  const iconsList = readCodepoints(codepointsPath)
+  const iconsAmount = iconsList.length
+
+  // Create map for icons with additional metadata
+  const icons = {}
+
+  // Download all icons
+  await utils.download('icons     ', iconsUrl, iconsPath)
```
michalochman: These extra spaces in label argument across this file are only for pretty printing to console?

packages/icon/scripts/utils.js
```diff
@@ -0,0 +1,203 @@
...
+// Build promisified functions
+const mkdir = promisify(fs.mkdirp)
+const writeFile = promisify(fs.writeFile)
+
+/**
+ * Clear line and display new text
+ *
+ * @param {string} str
+ */
+function writeLine (str) {
```
michalochman: This function looks more like an implementation of `overwriteLine`. `writeln` functions usually don't clear current output and they add new line characters.

michalochman: Also, it looks like it's not used anyway?

michalochman:
Please resolve conflicts before I can merge.

packages/icon/scripts/load-material-icons.js
rangoo94: Yes, exactly

packages/icon/scripts/utils.js
rangoo94: You're right, it's not used now - I will remove it (as well as `readline` dependency)

PR #7, opened 2018-03-21
[UI-10] Add generator of packages under `npm run create` command

rangoo94:


scripts/generatePackage.js
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * Validate title
+ *
+ * @param {string} title
+ * @returns {string|boolean}
+ */
+function validateTitle (title) {
+  if (title.length === 0) {
+    return `Title can't be empty`
```
michalochman: Any particular reason for using template literals?

rangoo94: Left after using `${title}` inside, will change it

PR #11, opened 2018-03-22
Feature/UI 10 migrate button

PiotrGrzechnik:


packages/button/README.md
```diff
@@ -0,0 +1,40 @@
...
+- `prop-types: ^15.6.1`
+- `react: ^16.2.0`
+- `react-dom: ^16.2.0`
+
+These packages are required by `@talixo/button`, but you have to install them manually,
+to avoid having different versions of these in your application.
+
+## Supported props
+
+Additionally, it handles some differently:
```
rangoo94: Something should be there? :)

packages/button/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Property name | Type      | Default | Description                    
+--------------|-----------|:-------:|--------------------------------
+className     | string    | n/a     | additional class name passed to wrapper
+children      | node      | n/a     | all button below
+color         | string    | n/a     | button color
```
rangoo94: For now `color` is okey, but we should rename it to use behaviors instead of colors

packages/button/README.md
```diff
@@ -0,0 +1,40 @@
...
+disabled      | boolean   | n/a     | enabled or disabled button
+id            | string    | n/a     | button id
```
rangoo94: Actually, these fields are not required, because they should be just passed

packages/button/src/Button.js
```diff
@@ -0,0 +1,80 @@
...
+ * @param {*} [props.onClick]
+ * @param {*} [props.size]
+ * @param {*} [props.style]
+ * @param {*} [props.variant]
+ * @returns {React.Element}
+ */
+function Button (props) {
+  const { children, className, color, disabled, id, onClick, size, style, variant, ...passedProps } = props
+
+  const mappedChildren = React.Children.map(children, child => {
```
rangoo94:
```js
const elements = React.Children.map(children, child => (
  typeof child === 'string' ? <span>{child}</span> : child
))
```

But do we really need to wrap it into `<span>`? AFAIR it was because of problems with icons inside

packages/button/src/Button.js
```diff
@@ -0,0 +1,80 @@
...
+  })
+  return (
+    <button
+      className={cls(name, className, {
+        [`${name}--${color}`]: color !== undefined,
+        [`${name}--${size}`]: size !== undefined,
+        [`${name}--${variant}`]: variant !== undefined
+      })}
+      style={style}
+      id={id}
```
rangoo94: These props are just passed

packages/button/stories.js
```diff
@@ -0,0 +1,74 @@
...
+    <span>Black</span>
+  </Button>
+))
+addStory('black-small', readme, () => (
+  <Button color='black' size='small'>
+    <span>Black</span>
+  </Button>
+))
+addStory('black-full-width', readme, () => (
+  <Button color='black' variant='full-width'>
```
rangoo94: Isn't `full-width` a size as well?

packages/button/styles/theming.sass
```diff
@@ -0,0 +1,59 @@
+@import './config'
+
+@include module('button')
+
+  &:hover:not(:disabled):not(&--black):not(&--ghost):not(&--full-width)
```
rangoo94: biiiig one :) you . could reverse this rule

packages/button/tests/Button.test.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+import Button from '../src/Button'
+import { shallow } from 'enzyme'
+
+describe('<Button />', () => {
+  it('renders children correctly', () => {
+    const wrapper = shallow(
+      <Button>
```
rangoo94: It will be worth to check button with different props as well

packages/button/styles/geometry.sass
```diff
@@ -0,0 +1,45 @@
...
+  border-color: $color-orange-talixo
+  font-size: $font-size-m
+  line-height: grid-unit(3)
+  cursor: pointer
+  box-shadow: $shadow-m rgba(0, 0, 0, 0.25)
+  color: $color-white
+  outline: inherit !important
+
+  > *
+    margin: 0 grid-unit(0.5)
```
rangoo94: Why it needs this margin?

packages/button/styles/theming.sass
rangoo94: actually, these rules or need needed

packages/button/styles/geometry.sass
PiotrGrzechnik: It's useful to separete elements inside the button (for ex. icon or span).

packages/button/stories.js
PiotrGrzechnik: After the discussion, we decided that It should stay as a variant because it has also different border-radius.

PR #13, opened 2018-03-22
Add possibility to create controlled stories with good source code shown

rangoo94:
As @maciek15 have shown and proven, there may be a problem with showing source code of components when we need to control them. Right now it ends up with creating wrapping stateful component which manages stuff inside. In source code section in Storybook it's visible as `<App />` without anything from `render()` method.

This Pull Request is implementing another, almost transparent, a method of creating such controlled components, with a simple syntax - you can see an example in README file.

### DIFF
--- a/packages/commons/story/createController.js
+++ b/packages/commons/story/createController.js
@@ -2,7 +2,13 @@ import React from 'react'
 
 import { action } from '@storybook/addon-actions'
 
+/**
+ * Wrapping React component to handle state changes
+ */
 class Controller extends React.Component {
+  /**
+   * @param {object|{ register: function, initialState: object, render: function }} props
+   */
   constructor (props) {
     super(props)
     this.state = props.initialState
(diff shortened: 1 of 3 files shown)


rangoo94:
Let me show an example:

```js
// ./src/Text.js
import React from 'react'
import PropTypes from 'prop-types'

function Text (props) {
  const { text, style, onClick } = props

  return (
    <div style={style}>
      {text}
      <button onClick={() => onClick()}>Increase</button>
    </div>
  )
}

Text.propTypes = {
  /** Some styles */
  style: PropTypes.object,
  /** Text to show */
  text: PropTypes.string,
  /** Handler */
  onClick: PropTypes.func
}

Text.defaultProps = {
  text: 10
}

export default Text


// ./stories.js
import React from 'react'

import Text from './src/Text'

import { createStoriesFactory } from '@talixo/commons/story'

const addStory = createStoriesFactory('Text', module)

function render (setState, state = {}) {
  return (
    <Text
      text='Just testing controlled stuff'
      style={{ fontSize: state.size }}
      onClick={() => setState({ size: state.size + 1 })}
    />
  )
}

function getInitialState () {
  return { size: 10 }
}

addStory.controlled('yuppi, controlled!', 'some description', render, getInitialState)
```
zrzut ekranu 2018-03-22 o 16 08 47

packages/commons/story/createController.js
```diff
@@ -0,0 +1,76 @@
...
+ */
+function createController (func, getInitialState = () => ({})) {
+  const state = getInitialState() || {}
+
+  // Create list of register controllers to update
+  const components = []
+
+  // Build some required functions
+  const register = e => components.push(e)
+  const update = () => components.filter(c => c.isMounted()).forEach(c => c.forceUpdate())
```
michalochman: `isMounted` is not safe, may be removed from react any time. It would be better to create `unregister` hook that components can call before they are unmounted and move `register` to `componentDidMount`

PR #14, opened 2018-03-22
[UI-10] Migrate configuration variables into shared package

rangoo94:


rangoo94:
It's just for now - we will need to adjust all stuff next week, to not make breaking changes anymore. For now it's in pixels, because AFAIR @stuniel and @maciek15 already made PRs with these variables in pixels

PR #17, opened 2018-03-23
[UI-10] Build new responsive Grid with special possibilities

rangoo94:


michalochman:
One thing not to forget is to use `em`s for breakpoints, see https://zellwk.com/blog/media-query-units/

rangoo94:
Breakpoints, for now, are just set to see if the grid is working, we have to set them up basing on either `em`s or our `dp` unit (`grid-unit`)

michalochman:
Sure, this was more like note to self.

PR #18, opened 2018-03-23
Migrate Tooltip component + Fix styles

stuniel:


packages/tooltip/README.md
```diff
@@ -0,0 +1,39 @@
...
+- `react-dom: ^16.2.0`
+
+These packages are required by `@talixo/tooltip`, but you have to install them manually,
+to avoid having different versions of these in your application.
+
+## Supported props
+
+Property name | Type                               | Default | Description                    
+--------------|------------------------------------|:-------:|-----------------------
+children      |	node	                             | n/a     | Tooltipped elements
```
rangoo94: Tab characters breaks it :)

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,194 @@
...
+
+    return (
+      <div
+        className={`${name}-hover`}
+        onClick={this.handleMouseClick}
+        onMouseEnter={this.handleMouseEnter}
+        onMouseLeave={this.handleMouseLeave}
+        onMouseOver={this.handleMouseOver}
+      >
+        <div ref={node => (this.el = node)} style={{ display: 'inline-block' }} className={`${name}-wrapper`}>
```
rangoo94: Could you extract this `ref` into method?

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,194 @@
...
+ * @param {string} [props.children]
+ * @param {string} [props.className]
+ * @param {string} [props.color]
+ * @param {string} [props.fade]
+ * @param {string} [props.fadeTime]
+ * @param {string} [props.isOpen]
+ * @param {string} [props.style]
+ * @param {string} [props.position]
+ * @param {string} [props.render]
+ * @returns {React.Element}
```
rangoo94: When you use `class` these params are not correct. It should be passed to `constructor`; `@returns` could be passed to `render`

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,194 @@
...
+  componentWillReceiveProps (nextProps) {
+    if (this.props.isOpen !== nextProps.isOpen) this.setState({ isOpen: nextProps.isOpen })
+  }
+
+  updatePosition () {
+    const { top, left } = getPositionNearElement(this.el, this.props.position)
+    this.setState({ top, left })
+  }
+
+  handleMouseEnter () {
```
rangoo94: Do you need both `mouseenter and `mouseover` events?

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,194 @@
...
+    if (!this.state.clicked && !this.state.isOpen) this.setState({ isOpen: true })
+  }
+
+  handleMouseClick () {
+    if (!_.isUndefined(this.props.isOpen)) return
+    this.setState({ clicked: true, isOpen: false })
+  }
+
+  render () {
+    const {
```
rangoo94: It could be smaller :)

packages/tooltip/src/Tooltip.js
rangoo94: When you use `class` these params are not correct. It should be passed to `constructor`; `@returns` could be passed to `render`

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,194 @@
...
+      rootNode,
+      style
+    } = this.props
+
+    const defaultFadeTime = 600
+    const transition = fade ? { transition: `opacity ${fadeTime || defaultFadeTime}ms` } : null
+
+    return (
+      <div
+        className={`${name}-hover`}
```
rangoo94: `--hover` may be better (for BEM)

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,194 @@
...
+  fade: PropTypes.bool,
+
+  /** Fade time */
+  fadeTime: PropTypes.number,
+
+  /** Controls whether tooltip is open */
+  isOpen: PropTypes.bool,
+
+  /** Tooltip position */
+  position: PropTypes.oneOf(['left', 'right', 'top', 'bottom']).isRequired,
```
rangoo94: Couldn't be `top` as default? It's pretty common

packages/tooltip/src/Tooltip.js
rangoo94: Do you need both `mouseenter` and `mouseover` events?

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,194 @@
...
+
+  /** Renders toolyip content */
```
rangoo94: typo

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,194 @@
...
+import { CSSTransition, TransitionGroup } from 'react-transition-group'
+
+import { getPositionNearElement } from '../utils/position'
+import { prefix } from '@talixo/commons'
+
+const name = prefix('tooltip')
+
+class Portal extends React.Component {
+  constructor (props) {
+    super(props)
```
rangoo94: It's not handling root change

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,194 @@
...
+  fadeTime: PropTypes.number,
+
+  /** Controls whether tooltip is open */
+  isOpen: PropTypes.bool,
+
+  /** Tooltip position */
+  position: PropTypes.oneOf(['left', 'right', 'top', 'bottom']).isRequired,
+
+  /** Renders toolyip content */
+  render: PropTypes.func,
```
rangoo94: Right now `render` is required

packages/tooltip/src/Tooltip.js
rangoo94: typo

packages/tooltip/utils/position.js
```diff
@@ -0,0 +1,31 @@
+export const getPositionNearElement = (element, position) => {
+  const boundingClientRect = element.getBoundingClientRect()
```
rangoo94: What if position of element changes? Right now no events are checked

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,215 @@
...
+import { getPositionNearElement } from '../utils/position'
+import { prefix } from '@talixo/commons'
+
+const name = prefix('tooltip')
+
+class Portal extends React.Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      root: this.props.root || document.querySelector('body')
```
rangoo94: Do you really need to keep it in state?

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,215 @@
...
+
+  handleMouseClick () {
+    if (!_.isUndefined(this.props.isOpen)) return
+    this.setState({ clicked: true, isOpen: false })
+  }
+
+  setRef (node, child) {
+    this.el = node
+    const {ref} = child
+    if (typeof ref === 'function') {
```
rangoo94: Why do you need that?

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,215 @@
...
+    return (
+      <div
+        className={`${name}-hover`}
+        onClick={this.handleMouseClick}
+        onMouseEnter={this.handleMouseEnter}
+        onMouseLeave={this.handleMouseLeave}
+        onMouseOver={this.handleMouseOver}
+      >
+        {React.Children.map(this.props.children, child =>
+          React.cloneElement(React.Children.only(child), {
```
rangoo94: When you are using `React.Children.map` isn't `child` always the only child there?

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,215 @@
...
+
+    return (
+      <div
+        className={`${name}-hover`}
+        onClick={this.handleMouseClick}
+        onMouseEnter={this.handleMouseEnter}
+        onMouseLeave={this.handleMouseLeave}
+        onMouseOver={this.handleMouseOver}
+      >
+        {React.Children.map(this.props.children, child =>
```
rangoo94: Could you extract some of these above for clearer code?

packages/tooltip/stories.js
```diff
@@ -0,0 +1,105 @@
...
+        backgroundColor: '#eee',
+        padding: '1rem',
+        cursor: 'pointer'
+      }}
+    >
+      Tooltip right
+    </span>
+  </Tooltip>
+))
+addStory('controlled', readme, () => <ControlledTooltip />)
```
rangoo94: could you use `addStory.controlled` to have problem source code visible in Storybook?

packages/tooltip/styles/geometry.sass
```diff
@@ -0,0 +1,92 @@
+@import './config'
+
+@include module('tooltip')
+  position: absolute
+  display: inline-block
+  padding: grid-unit(1) grid-unit(1.5)
+  border: $border-width solid
```
rangoo94: `border: $border-width solid $color-orange-talixo` and you can get rid of `border-color` below

packages/tooltip/styles/geometry.sass
```diff
@@ -0,0 +1,92 @@
...
+    opacity: 1
+
+  &-fade-exit-active
+    opacity: 0.01
+
+  &--right
+    transform: translateY(-50%) translateX(grid-unit(1))
+
+    &:before
+      transform: translateY(-50%) translateX(50%) rotate(-45deg) skew(20deg, 20deg)
```
rangoo94: What is this `skew` for?

rangoo94: As I understand you are creating triangles - could you use it with just standard border triangles?
http://apps.eky.hk/css-triangle-generator/

packages/tooltip/src/Tooltip.js
```diff
@@ -0,0 +1,213 @@
...
+                    [`${name}--${color}`]: color !== undefined,
+                    [`${name}--${position}`]: position !== undefined
+                  })}
+                  style={{
+                    top: this.state.top,
+                    left: this.state.left,
+                    ...transition,
+                    ...style
+                  }}
+                >
```
maciek15: `className` and `style` could be extracted above

packages/tooltip/README.md
```diff
@@ -0,0 +1,39 @@
...
+- `react-dom: ^16.2.0`
+
+These packages are required by `@talixo/tooltip`, but you have to install them manually,
+to avoid having different versions of these in your application.
+
+## Supported props
+
+Property name | Type                               | Default | Description
+--------------|------------------------------------|:-------:|-----------------------
+children      |	node	                             | n/a     | Tooltipped elements
```
michalochman: Tab characters

PR #21, opened 2018-03-23
Migrate Tab component

stuniel:


packages/tab/src/Tab.js
```diff
@@ -0,0 +1,62 @@
...
+  /** Tab content */
+  children: PropTypes.node,
+
+  /** Additional class name */
+  className: PropTypes.string,
+
+  /** Id passed to tab */
+  id: PropTypes.string,
+
+  /** Function that runs whe tab is clicked */
```
rangoo94: typo

packages/tab/README.md
```diff
@@ -0,0 +1,37 @@
...
+These packages are required by `@talixo/tab`, but you have to install them manually,
+to avoid having different versions of these in your application.
+
+## Supported props
+
+It allows any props which are allowed for `div`. Additionally, it handles some differently:
+
+Property name | Type      | Default | Description                    
+--------------|-----------|:-------:|--------------------------------
+active        |	bool      |	"false"	| Active state
```
rangoo94: Tab characters in tab :)

packages/tab/src/Tab.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {string} [props.active]
+ * @param {string} [props.children]
+ * @param {string} [props.className]
+ * @param {string} [props.id]
+ * @param {string} [props.onClick]
+ * @param {string} [props.style]
+ * @returns {React.Element}
+ */
+const Tab = ({ active, children, className, id, onClick, style }) => {
+  const mappedChildren = React.Children.map(children, child => {
```
rangoo94: Do we really need to wrap text with `span`?

packages/tab/stories.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+// Stories
+
+addStory('default', readme, () => (
+  <Tab>Home</Tab>
+))
+addStory('group', readme, () => (
+  <TabContainer>
```
rangoo94: Actually, as I see, we've got few components which are the same (but has different styling):

- `Breadcrumbs`
- `Pagination`
- `TabContainer`

The truth is, that all of them are just `Navigation` - we should consider that

packages/tab/styles/geometry.sass
```diff
@@ -0,0 +1,21 @@
...
+  padding: grid-unit(1) grid-unit(1.5)
+  border-width: $border-width $border-width 0 $border-width
+  border-style: solid
+  border-radius: $border-radius-m $border-radius-m 0 0
+  box-shadow: $shadow-m rgba(0, 0, 0, 0.25)
+  font-size: $font-size-s
+  cursor: pointer
+
+  > *
+    margin: 0 0.5 * $base-size
```
rangoo94: `grid-unit/dp`? :)

packages/tab/stories.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  render () {
+    const { children } = this.props
+    return (
+      <div>
+        {React.Children.map(children, child =>
+          React.cloneElement(child, {
+            active: child.props.id === this.state.activeTab,
+            onClick: () => this.setState({ activeTab: child.props.id })
```
rangoo94: It's overriding `Tab` onClick event - try to make test:
```js
it('calls onClick when clicked', () => {
  const onClick = jest.fn()
  const wrapper = mount(
  <TabContainer>
    <Tab onClick={onClick}>Tab</Tab>
  </TabContainer>
)

  wrapper.find(`.${name}`).simulate('click')
  expect(onClick).toHaveBeenCalledTimes(1)
  wrapper.unmount()
})
```

Because it will fail now. Additionally, `TabContainer` shouldn't have state - why it should change tab without a knowledge about their behavior?

rangoo94: So it will be good to both remove cloning element, state and writing this test for regression :)

packages/tab/README.md
```diff
@@ -0,0 +1,37 @@
...
+It allows any props which are allowed for `div`. Additionally, it handles some differently:
+
+Property name | Type      | Default | Description
+--------------|-----------|:-------:|--------------------------------
+active        |	bool      |	"false"	| Active state
+children      |	node      | n/a     | Tab content
+className     |	string    | n/a     | Additional class name
+id            |	string    | n/a     | Id passed to tab
+onClick       |	func      | n/a     | Function that runs when tab is clicked
+style         |		        | n/a     | Additional styles passed to the tab
```
rangoo94: Tabs

packages/tab/src/Tab.js
```diff
@@ -0,0 +1,59 @@
...
+ * @param {object} [props.style]
+ * @returns {React.Element}
+ */
+const Tab = ({ active, children, className, id, onClick, style }) => {
+  return (
+    <div
+      className={cls(name, className, {
+        active: active !== false
+      })}
+      style={style}
```
rangoo94: Could you just pass rest of `props` here (inc. `style`, `id`, `onClick`)

packages/tab/src/Tab.js
```diff
@@ -0,0 +1,59 @@
...
+  active: PropTypes.bool,
+
+  /** Tab content */
+  children: PropTypes.node,
+
+  /** Additional class name */
+  className: PropTypes.string,
+
+  /** Id passed to tab */
+  id: PropTypes.string,
```
rangoo94: As before, these `props` are not anything special

packages/tab/src/Tab.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cls from 'classnames'
+import { prefix } from '@talixo/shared'
+
+const name = prefix('tab')
```
PiotrGrzechnik: Please change `name` to `moduleName`.

packages/tab/src/Tab.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param {string} [props.className]
+ * @param {string} [props.id]
+ * @param {*} [props.onClick]
+ * @param {object} [props.style]
+ * @returns {React.Element}
+ */
+const Tab = ({ active, children, className, ...passedProps }) => {
+  return (
+    <div
+      className={cls(name, className, {
```
PiotrGrzechnik: The same as above.

packages/tab/styles/geometry.sass
```diff
@@ -0,0 +1,22 @@
+@import './config'
+
+@include module('tab')
+  // display: inline-block
```
PiotrGrzechnik: If it's not necessary, please remove this comment.

packages/tab/src/Tab.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {string} [props.active]
+ * @param {*} [props.children]
+ * @param {string} [props.className]
+ * @returns {React.Element}
+ */
+const Tab = ({ active, children, className, ...passedProps }) => {
+  return (
+    <div
+      className={cls(moduleName, className, {
+        active: active !== false
```
maciek15: Shouldn't class active be added only when `active === true` ?
Also `.active should be prefixed`

packages/tab/styles/geometry.sass
```diff
@@ -0,0 +1,23 @@
...
+    margin-left: 1px
+
+  > *
+    margin: 0 dp(0.5)
+
+  &:active
+    font-weight: $font-weight-semibold
+
+  &.active
+    font-weight: $font-weight-semibold
```
maciek15: Class `.active` should be prefixed in case there was other class with the same name

PR #22, opened 2018-03-23
Migrate Notification component

stuniel:


packages/notification/src/Notifications.js
```diff
@@ -0,0 +1,73 @@
...
+class Notifications extends React.Component {
+  constructor () {
+    super()
+    this.state = {
+      items: []
+    }
+  }
+
+  componentDidMount () {
+    setTimeout(() => this.setState({items: Array.isArray(this.props.children) ? this.props.children : [this.props.children]}), 1000)
```
rangoo94: - Could you format it? :)
- You can use `concat`
- It may be a fragment
- Why after 1 second?
- When component will receive props, these should be updated - that means that this component shouldn't worry about hiding them, it should emit event only.
- Additionally, this timeout should be cleared on unmount:

```js
componentDidMount () {
  this.timeout = setTimeout(() => this.setState({
    items: [].concat(this.props.children)
  })
}

componentWillUnmount () {
  clearTimeout(this.timeout)
}
```

packages/notification/src/Notifications.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  handleRemove (i) {
+    let newItems = this.state.items.slice()
```
rangoo94: Or eventually:
```js
const { items } = this.state

this.setState({
  items: items.filter((x, index) => index !== i)
})
```

packages/notification/src/Notifications.js
```diff
@@ -0,0 +1,73 @@
...
+    newItems.splice(i, 1)
+    this.setState({ items: newItems })
+  }
+
+  render () {
+    const { className, style } = this.props
```
rangoo94: Should style be passed to all notifications?

rangoo94: Class too?

rangoo94: When you use composition (and you do), actually you could add them from above. This stuff should go either to `container` or `wrapper`.

packages/notification/src/Notifications.js
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <div className={`${name}-container`}>
+        <div className={`${name}-wrapper`}>
+          <TransitionGroup className={name}>
+            {this.state.items.map((child, i) => (
+              <Fade key={child.props.id}>
+                {React.cloneElement(child, {
+                  className,
+                  style,
+                  onClick: () => this.handleRemove(i)
```
rangoo94: Should notification be removed after click anywhere?

packages/notification/styles/geometry.sass
```diff
@@ -0,0 +1,80 @@
...
+  width: grid-unit(50)
+  display: flex
+  flex-wrap: wrap
+  flex-direction: column
+
+  > *
+    margin: grid-unit(1) grid-unit(2)
+
+  &-container
+    position: fixed
```
rangoo94: Nothing below will be clickable then

packages/notification/styles/geometry.sass
```diff
@@ -0,0 +1,80 @@
...
+    top: 0
+    left: 0
+    height: 100vh
```
rangoo94: you don't have to use viewport units when using `position: fixed`, `width: 100%; height: 100%` is enough

packages/notification/styles/geometry.sass
```diff
@@ -0,0 +1,80 @@
...
+    width: 100vw
+
+  &-wrapper
+    position: absolute
+    height: grid-unit(19)
```
rangoo94: Should it be `19` or dynamic?

packages/notification/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+These packages are required by `@talixo/notification`, but you have to install them manually,
+to avoid having different versions of these in your application.
+
+## Supported props
+
+Property name | Type      | Default | Description
+--------------|-----------|:-------:|------------------------------------------------
+className     | string    | n/a     | additional class name passed to wrapper
+children      |	node      | n/a     |	Notification content
```
rangoo94: Tab characters ;)

packages/notification/index.js
```diff
@@ -0,0 +1,3 @@
+import Notification from './src/Notification'
+
+export default Notification
```
rangoo94: Shouldn't you export also `NotificationsList`?

packages/notification/src/Notification.js
```diff
@@ -0,0 +1,58 @@
...
+ * @param {string} [props.className]
+ * @param {string} [props.handleRemove]
+ * @param {string} [props.style]
+ * @param {string} [props.variant]
+ * @returns {React.Element}
+ */
+const Notification = props => {
+  const { children, className, handleRemove, style, variant } = props
+  const mappedChildren = React.Children.map(children, child => {
+    return typeof child === 'string' ? <span>{child}</span> : child
```
rangoo94: So we keep it in `span`?

packages/notification/src/Notification.js
```diff
@@ -0,0 +1,58 @@
...
+  })
+  return (
+    <div
+      className={cls(name, className, {
```
rangoo94: Could you extract it above?

rangoo94:
Also, please resolve conflicts

packages/notification/src/Notification.js
```diff
@@ -0,0 +1,54 @@
...
+ * @param {string} [props.handleRemove]
+ * @param {string} [props.style]
+ * @param {string} [props.variant]
+ * @returns {React.Element}
+ */
+const Notification = props => {
+  const { children, className, handleRemove, style, variant, ...passedProps } = props
+  const classNames = cls(name, className, {
+    [`${name}--${variant}`]: variant !== undefined
+  })
```
maciek15: you can modify this condition to `variant && variant length > 0` to avoid error when null is passed and additional class when empty string is passed

packages/notification/src/Notification.js
```diff
@@ -0,0 +1,54 @@
...
+
+  /** Function that runs when close button is clicked */
+  handleRemove: PropTypes.func,
+
+  /** Additional styles passed to notification */
+  style: PropTypes.object,
+
+  /** Notification variant */
+  variant: PropTypes.string
+}
```
maciek15: I think 'variant' can be enumerated here using `PropTypes.oneOf([ ])`, because only 5 strings can actually modify notifications

packages/notification/src/Notification.js
```diff
@@ -0,0 +1,54 @@
...
+ * @returns {React.Element}
+ */
+const Notification = props => {
+  const { children, className, handleRemove, style, variant, ...passedProps } = props
+  const classNames = cls(name, className, {
+    [`${name}--${variant}`]: variant !== undefined
+  })
+
+  return (
+    <div className={classNames} style={style} {...passedProps}>
```
maciek15: className and style here is passed just like any other prop inside `{ ...passedProps }`. You can just mention it inside readme that this component allows any props for divs

packages/notification/styles/geometry.sass
```diff
@@ -0,0 +1,94 @@
...
+  align-items: flex-start
+  width: 100%
+  padding: dp(2) dp(4)
+  border: $border-width solid
+  border-radius: $border-radius-m
+  box-shadow: $shadow-l rgba(0, 0, 0, 0.25)
+  font-size: $font-size-m
+  overflow: hidden
+
+  & .close
```
maciek15: `.close` should be prefixed

packages/notification/styles/geometry.sass
```diff
@@ -0,0 +1,94 @@
...
+  position: relative
+  width: dp(50)
+  display: flex
+  flex-wrap: wrap
+  flex-direction: column
+
+  @include module('notification')
+    margin: $marginY $marginX
+
+  &-container
```
maciek15: As we spoke with @rangoo94 containers and wrappers should also be prefixed like elements `__container`.
@rangoo94, can you confirm?

packages/notification/styles/geometry.sass
```diff
@@ -0,0 +1,94 @@
...
+    position: fixed
+    top: 0
+    left: 0
+    height: 100%
+    width: 100%
+
+  &-wrapper
```
maciek15: Same as before

packages/notification/README.md
```diff
@@ -21,15 +22,37 @@ to avoid having different versions of these in your application.
...
+
+variant       |
+--------------|
+'primary'     |
+'success'     |
+'error'       |
+'warning'     |
+'information' |
+
+### NotificationsList
```
michalochman: I would suggest renaming this component to `NotificationList`

PR #27, opened 2018-03-27
Migrate sidebar into UI Kit

rangoo94:
Additionally, now it allows nesting of panels and sidebars.


packages/sidebar/index.js
```diff
@@ -0,0 +1,9 @@
+import Sidebar from './src/Sidebar'
+import Element from './src/Element'
+import Panel from './src/Panel'
+
+export {
+  Sidebar,
+  Element,
+  Panel
```
michalochman: `Element` and `Panel` are ambiguous names, multiple packages can export such components. 

I'm hesitant to transfer the responsibility of renaming this to e.g. `import { Element as SidebarElement } from @talixo/sidebar`, since it may create lack of consistency in naming convention. Wouldn't it be better to rename it here to `SidebarElement` or something less ambiguous?

packages/sidebar/src/Element.js
```diff
@@ -0,0 +1,203 @@
...
+  constructor (props) {
+    super(props)
+
+    this.handleClick = this.handleClick.bind(this)
+    this.handleClose = this.handleClose.bind(this)
+    this.saveRef = this.saveRef.bind(this)
+
+    this.node = null
+    this.state = {
+      opened: false
```
michalochman: `open`?

packages/sidebar/index.js
michalochman: `Element` and `Panel` are ambiguous names, multiple packages can export such components. 

I'm hesitant to transfer the responsibility of renaming this to e.g. `import { Element as SidebarElement } from @talixo/sidebar`, since it may create lack of consistency in naming convention. Wouldn't it be better to rename it here to `SidebarElement` or something less ambiguous?

packages/sidebar/src/Element.js
michalochman: `open`?

packages/sidebar/src/Element.js
```diff
@@ -0,0 +1,203 @@
...
+ * @property {*} [props.children]
+ * @property {string} props.icon
+ * @property {string|*} props.label
+ * @property {boolean} props.active
+ *
+ * @property {object} state
+ * @property {boolean} state.opened
+ *
+ * @property {Element} node  reference to DOM element of Sidebar
+ */
```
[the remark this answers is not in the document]
rangoo94: Actually, there could be `@class` property instead, but you're right ;)

packages/sidebar/index.js
rangoo94: I was thinking about it and had doubts as you did, so it may be `SidebarElement`, or maybe anyone here has any better idea for its name?

Otherwise we may also think about proposition which you gave previously:

```js
import Sidebar from '@talixo/sidebar'

function App () {
  return (
    <Sidebar>
      <Sidebar.Element icon="home" label="Dashboard">
        <Sidebar.Panel name="Blabla">
          Blablabla
        </Sidebar.Panel>
      </Sidebar.Element>
    </Sidebar>
  )
}
```

michalochman: I prefer `Sidebar.` "namespace" tbh

packages/sidebar/src/Element.js
rangoo94: `isOpen` as in others

PR #30, opened 2018-03-28
Build Table components

rangoo94:
- Unfortunately, this PR includes typography from another Pull Request

The component itself works properly both in LTR and RTL directions. Additionally, we will create `DataTable` component/package later, which will have responsibility of putting data into this `Table` structure.

### DIFF
--- a/packages/table/styles/geometry.sass
+++ b/packages/table/styles/geometry.sass
@@ -1,0 +1,107 @@
+@import './config'
+
+@include module('table')
+  width: 100%
+  border-spacing: 0
+  border: 1px solid $table-border-color
+
+  &, td, th
+    box-sizing: border-box
+
+  td, th
+    text-align: left
+    padding: dp(1) dp(1.5)
+
+  th
+    font-weight: $table-head-text-weight
+    color: $table-head-text-color
+    border-bottom: dp(0.5) solid $table-border-color
+
+  td
+    color: $table-row-text-color
+    background: $table-row-background-color
+
+  tfoot td
+    border-top: dp(0.5) solid $table-border-color
+
+  &--fixed
+    table-layout: fixed
+
+    #{prefix('table', 'actions')}
+      width: auto
+      white-space: normal
+
+  &--condensed
+    td, th
+      padding: dp(0.5) dp(1)
+
+    #{prefix('table', 'action')}
+      flex-direction: row
+      padding: dp(0.5) dp(1.5)
+
+      > .tai
+        margin-right: dp(0.5)
+
+@include module('table', 'action')
+  box-sizing: border-box
+  display: inline-flex
+  height: 100%
+  outline: none
+  background: transparent
+  align-items: center
+  flex-direction: column
+  text-align: center
+  text-decoration: none
+  cursor: pointer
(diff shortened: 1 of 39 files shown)


packages/table/styles/geometry.sass
```diff
@@ -0,0 +1,107 @@
...
+@include module('table')
+  width: 100%
+  border-spacing: 0
+  border: 1px solid $table-border-color
+
+  &, td, th
+    box-sizing: border-box
+
+  td, th
+    text-align: left
```
optik: Use `text-align: start` (optionally with fallback to `text-align: left`)
OR
Consider allowing browser to inherit text alignment for "normal" cells from parent.

packages/table/styles/geometry.sass
```diff
@@ -0,0 +1,107 @@
...
+  &--condensed
+    td, th
+      padding: dp(0.5) dp(1)
+
+    #{prefix('table', 'action')}
+      flex-direction: row
+      padding: dp(0.5) dp(1.5)
+
+      > .tai
+        margin-right: dp(0.5)
```
optik: Ou could avoid adding additional `[dir=rtl]` rule by adding spacing symmetrically with some sort of compensation in the container.

packages/table/styles/geometry.sass
```diff
@@ -0,0 +1,107 @@
...
+  @include module('table', 'action')
+    border-left-width: 0
+    border-right-width: 1px
+
+  @include module('table', 'actions')
+    &:last-child
+      padding-left: 0
+      padding-right: dp(1.5)
+
+  #{prefix('table')}--condensed #{prefix('table', 'action')} > .tai,
```
optik: Could be avoided (see comment re icon spacing)

packages/table/styles/geometry.sass
```diff
@@ -0,0 +1,107 @@
...
+        margin-right: dp(0.5)
+
+      &:last-child
+        border-bottom-width: 0
+
+
+[dir=rtl]
+  @include module('table')
+    td, th
+      text-align: right
```
optik: See first comment re text-align. Additionally, we may come into rule precendence issues in RTL when styling individual tables/cells.

packages/table/styles/geometry.sass
rangoo94: I am not sure if we should inherit alignment from the parent, but I will change it for now to `text-align: start` with fallback to `left/right` depending on direction, thanks!

packages/table/styles/geometry.sass
rangoo94: We couldn't avoid it for the icon, as it will create additional space in the parent container. Otherwise, when we will remove this `padding` from the parent, we may come into problems with displaying text without icons

packages/table/styles/geometry.sass
rangoo94: You're right, I removed this rule - IE10 is not our best target, and using this rule over here may cause some problems.

rangoo94: You're right, I will just keep `text-align: start` in basic styling and we will come into small bugs in <IE10.

PR #33, opened 2018-04-03
[UI-1] Add new CodeClimate checks, remove unused libraries

rangoo94:


michalochman:
Where is `markdownlint-cli` used?

rangoo94:
Sorry, I thought that I removed it, thanks for check!

rangoo94:
I added an issue to CodeClimate duplication plugin:
https://github.com/codeclimate/codeclimate-duplication/issues/306

There is a better handler for duplication detection on `beta` channel, but it's not working with newer ES6. I will prepare bugfix for them on the evening, so probably we could switch to `beta` then.

Issue #32, opened 2018-04-03
Popover - warning in console (wrong propType)


maciek15:
There is a warning in console in storybook: Invalid prop `fade` of type `string` supplied to `Popover`, expected `boolean`

PiotrGrzechnik:
Please see http://jira.talixo.de/browse/UI-34

rangoo94:
@PiotrGrzechnik has updated Tooltip/Popover already so it's working correctly now.

PR #35, opened 2018-04-03
UI-34  Popover - warning in console (wrong propType) & change first story in popover.

PiotrGrzechnik:
It's resolution for #32.


packages/popover/stories.js
```diff
@@ -35,33 +35,32 @@ function render (setState, state) {
...
       <div>
-        <Button id='popover-bottom' onClick={() => togglePopover('popover-bottom')} color='primary'>
-          Popover default right
+        <Button id='popover-default' onClick={() => togglePopover('popover-default')} color='primary'>
+          Popover default
         </Button>
         <Popover
-          target='#popover-bottom'
-          isOpen={state.activeTarget === 'popover-bottom'}
+          target='#popover-default'
```
rangoo94: Is `target` required in `Popover`?

PiotrGrzechnik: Popover and button are independent of each other, so that's why we need use target.

rangoo94:
Popover will be merged with tooltip, so I am closing PR

PR #38, opened 2018-04-03
Feature/ui 11 set up typography and other basic styles

rangoo94:


packages/typography/src/Heading.js
```diff
@@ -0,0 +1,57 @@
...
+  const { className, level, secondary, children, ...passedProps } = props
+
+  // Calculate heading level
+  const desiredLevel = Math.max(1, Math.min(level, 6))
+
+  // Calculate children
+  const content = secondary ? (
+    <React.Fragment>
+      {children}
+      <small>{secondary}</small>
```
michalochman: Is this special handling of `secondary` necessary? The usage feels awkward. I'd just pass it directly as a part of `children`.

packages/typography/src/Lead.js
```diff
@@ -0,0 +1,36 @@
...
+
+const moduleName = prefix('lead')
+
+/**
+ * Component which represents article/page lead.
+ *
+ * @param {object} props
+ * @param {string} [props.className]
+ * @param {*} [props.children]
+ * @param {string|*} [props.secondary]
```
michalochman: This is neither defined nor used

packages/typography/stories.js
```diff
@@ -1,35 +1,153 @@
...
+    <Heading level={1} secondary='Secondary text'>h1. Header</Heading>
+    <Heading level={2} secondary='Secondary text'>h2. Header</Heading>
+    <Heading level={3} secondary='Secondary text'>h3. Header</Heading>
+    <Heading level={4} secondary='Secondary text'>h4. Header</Heading>
+    <Heading level={5} secondary='Secondary text'>h5. Header</Heading>
+    <Heading level={6} secondary='Secondary text'>h6. Header</Heading>
+
+    <Heading level={1}>h1. Header</Heading>
+    <Lead>
+      Lorem ipsum dolor sit amet enim. Etiam ullamcorper. Suspendisse a pellentesque dui, non felis.
```
michalochman: Should we consider creating `<Lipsum />` component for quick prototyping?

packages/typography/styles/geometry.sass
```diff
@@ -8,12 +11,29 @@ html
 html, body
   margin: 0
   padding: 0
+  -webkit-font-smoothing: antialiased
```
michalochman: Are you sure we want to apply it globally? I would suggest adding it through some kind of `.antialiased ` class. 

We should also add Firefox counterpart for OSX: `-moz-osx-font-smoothing: grayscale`.

packages/typography/styles/geometry.sass
```diff
@@ -8,12 +11,29 @@ html
...
+
+img
+  max-width: 100%
+
+// Anchors
+
+a
+  color: $color-ghost-text
+  text-decoration: none
+  border-bottom: 1px dotted $color-ghost-text
```
michalochman: Borders on anchors are problematic when these anchors have `block` children. I know this is against spec, but is a common use case. What's your take?

packages/shared/styles/config/base-unit.sass
```diff
@@ -1,7 +1,7 @@
 // Base unit
 
-$base-px: 8px
-$base-dp: 0.5rem // equivalent of 8px
+$base-px: 8px !default
```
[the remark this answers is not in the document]
rangoo94: As we are keeping `8px` grid I think that it's better to keep `$base-px` as the real base value

packages/typography/src/Heading.js
rangoo94: It's just an alternative syntax, which is more self-descriptive than using `<small>` inside

packages/typography/stories.js
rangoo94: More likely `lipsum` function, but we can use external libraries (i.e. `lorem-ipsum`). Anyway, it's used only in this typography story, so I don't think that it's worth to load it.

You can see in other stories in this file, that `Lorem ipsum` text can be mixed with some components inside, but better is to use real content which matches the context of the specific component.

packages/typography/styles/geometry.sass
rangoo94: Is there any reason to not use it globally? I am using these this way for few years and haven't found any downsides. It's something what makes text smoother on browsers which allow that, but anyway - it's optional, progressive enhancement.

packages/typography/styles/geometry.sass
rangoo94: `border` instead of `text-decoration` here is just a nice replacement (which looks slightly better).

The truth is that both `border` and `text-decoration` may be problematic. In case of anchors which are `block` elements, most probably they are separated components, so they should take care of itself by working with both of these options.

Anyway, of course, I see downsides there, so if you would prefer to have `text-decoration` instead I have nothing against.

packages/typography/src/Heading.js
michalochman: What if you wanted `tertiary`, `quaternary` and so on?

The result is that now you see `secondary` first, as it is a prop:

`<Heading level={1} secondary='Secondary text'>h1. Header</Heading>`

If you're worried about self-descriptiveness then maybe it's better to just extract `<small />` to a separate component, something like `<SecondaryText />`?

packages/typography/styles/geometry.sass
michalochman: I'm not against borders in general – only against setting them by default for all anchor elements.

packages/typography/styles/geometry.sass
michalochman: There are several problems:
1) It may be confusing when compared on different OS / Browser combinations
2) It mostly makes sense for light text on dark background
3) It's missing OSX / Firefox support (as mentioned above) – btw, shouldn't we use a tool like `autoprefixer`?

rangoo94: Yes, we should use `autoprefixer`, thanks for the reminder :) Anyway, I don't think that `autoprefixer` will handle that case, as `-moz-osx-font-smoothing: grayscale` is doing different thing than `-webkit-font-smoothing: antialiased`, despite fact that effect looks the same.

packages/typography/styles/geometry.sass
rangoo94: I was just asking about this specific (`global`) situation :) Anyway, I will just use `text-decoration` then.

packages/typography/src/Heading.js
rangoo94: I will just remove `secondary` then :)

Issue #41, opened 2018-04-04
Change rootNode type from string to node


stuniel:
String selector for `rootNode` prop is amiguous. It is much clearer to provide an actual node to the `root` prop.

```
<Portal root={this.props.rootNode}>
```

```
rootNode: PropTypes.node,
```

stuniel:
#42

stuniel:
Closes #39

PR #42, opened 2018-04-04
Unify prop names + Extract and refector Portal component

stuniel:


stuniel:
Closes #41

packages/modal/src/Modal.js
```diff
@@ -1,90 +1,53 @@
 import React from 'react'
-import ReactDOM from 'react-dom'
 import PropTypes from 'prop-types'
 import cls from 'classnames'
+import Portal from './Portal'
```
rangoo94: AFAIR Portal had to be extracted to separated package, am I right @michalochman ?

packages/modal/src/Modal.js
```diff
@@ -108,14 +71,14 @@ Modal.propTypes = {
   children: PropTypes.node,
 
   /** Controls whether modal is open */
-  isOpen: PropTypes.bool,
+  open: PropTypes.bool,
 
   /** Root element of modal */
-  root: PropTypes.object
+  attachTo: PropTypes.object
```
rangoo94: Good name :+1:

packages/modal/src/Portal.js
```diff
@@ -0,0 +1,42 @@
...
+ * @property {*} [props.children]
+ * @class
+ */
+class Portal extends React.Component {
+  constructor (props) {
+    super(props)
+    this.el = document.createElement('div')
+  }
+
+  componentDidMount () {
```
rangoo94: If we have proper `Portal` component maybe it's worth to handle a change of `attachTo` component?

packages/modal/src/Modal.js
michalochman: We decided to extract it with next PR.

packages/modal/src/Portal.js
rangoo94: Also, shouldn't we extract is as separated package as it's used in many others?

michalochman: Let's extract it after it's merged.

rangoo94:
@stuniel, could you resolve conflicts to merge it? ;)

packages/modal/src/Portal.js
rangoo94: As a Portal is here exposed in `Popover`, I will do it ;)

PR #45, opened 2018-04-05
[UI-35] Controlled components updated, does not show Control PropTypes

maciek15:
#### Task

- **Issue or task referred:** UI-35
- **Feature or Bugfix:** Bugfix

Changed two lines to overwrite `infoOptions.propTablesExcluded` with `excludedControllers`, fixes #44

PR #47, opened 2018-04-05
[UI-15] Add examples for existing components

maciek15:
#### Task

- **Issue or task referred:** UI-15

I've added more Storybook examples to components listed below. I did not apply any changes to components that will be refactored: Breadcrumbs, Popover, Tooltip. Additionally I've made some changes to some components styling.

#### Components

- Button (minimal margin to both sides)
- Badge (padding and default colors added)
- Checkbox
- List (default bullet added -> '-')
- Modal
- Notification (default style added)
- NumberInput
- Pill(default styling added)
- Radio (sizes added: 'small', large')
- Switcher
- TextInput

### DIFF
--- a/packages/badge/stories.js
+++ b/packages/badge/stories.js
@@ -24,7 +24,7 @@ addStory('default', readme, () =>
 addStory('inside button', readme, () =>
   <div>
     <Button>
-      Mail <Badge>10 unreaded</Badge>
+      Mail <Badge>10 unread</Badge>
     </Button>
     <Button color='primary'>
       Warnings <Badge>10 problems</Badge>
@@ -38,7 +38,7 @@ addStory('inside tab', readme, () =>
   <div>
     <Tab>
       Mail
-      <Badge>10 unreaded</Badge>
+      <Badge>10 unread</Badge>
     </Tab>
     <Tab>
       info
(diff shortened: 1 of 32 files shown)


rangoo94:
If component requires other component only for Storybook, it should be included in `devDependencies` section

packages/badge/stories.js
```diff
@@ -1,11 +1,52 @@
...
+    <Badge>1</Badge>{' '}
+    <Badge>10</Badge>{' '}
+    <Badge>100</Badge>{' '}
+    <Badge>Badge</Badge>
+  </div>
+)
+addStory('inside button', readme, () =>
+  <div>
+    <Button>
+      Mail <Badge>10 unreaded</Badge>
```
michalochman: unread ;)

packages/modal/stories.js
```diff
@@ -32,5 +33,9 @@ const getInitialState = () => {
     isOpen: false
   }
 }
+const options = {
+  propTypes: [ Modal, Button ],
+  propTablesExclude: [ Button ]
```
michalochman: Are you sure we need `Button` in both?

maciek15: Yes, it was required to display headers for PropTypes table of modal. But this problem is fixed in #45 so I will delete it now

PR #52, opened 2018-04-09
[UI-27] Handle formatters inside inputs

### LINKED ISSUE #43: Changing number input does not force onChange event
#### Information

- **Feature Request or Bug:** Bug
- **Components and versions affected:** NumberInput 1.0.0
- **Environment:** Ubuntu 16.04, Node 8.11

Changing number input does not force onChange event

#### How to reproduce?

- Pass logging function to onChange prop
- There is no logging information in console

#### What is expected behavior?

- onChange function should be fired when user changes input value

maciek15:
#### Task

- **Issue or task referred:** UI-27

Added possibility to display suffixes in number input. NumberInput is now a controlled component with inputValue in state. It also fixes #43 as both button and input itself return event and input value


packages/number-input/src/NumberInput.js
```diff
@@ -10,91 +10,142 @@ const moduleName = prefix('number-input')
...
+      [`${moduleName}--error`]: hasError
+    })
+
+    return (
+      <div className={wrapperClass} style={style}>
+
+        {
+          suffix
+            ? <div className={`${moduleName}__suffix-wrapper`}>
+              <span className={`${moduleName}__suffix-value`}>{inputValue}</span>
```
michalochman: Why there's no `<input />` rendered with `suffix`?

rangoo94:
It should use always `input` inside. Right now, when you provide suffix there is no way to set value from keyboard

packages/number-input/src/NumberInput.js
```diff
@@ -54,7 +54,7 @@ class NumberInput extends React.Component {
   onButtonClick (step, e) {
     const { maxValue, minValue, onClick } = this.props
     const { inputValue } = this.state
-    const newValue = !isNaN(inputValue) ? inputValue + step : 0
+    const newValue = !isNaN(parseInt(inputValue)) ? parseInt(inputValue) + step : 0
```
michalochman: Please always use radix (second argument to `parseInt`). Also, it would be nice to extract `parseInt` call, there's no reason to call it twice.

PR #56, opened 2018-04-12
Feature/UI-18 tooltip & popover as one component

PiotrGrzechnik:


rangoo94:
Tests failed; you're using `classnames` somewhere, instead of methods from `@talixo/shared` (especially `buildClassName`)

packages/tooltip/src/Tooltip.js
```diff
@@ -169,10 +173,10 @@ class Tooltip extends React.Component {
...
       <div
         className={`${moduleName}-hover`}
-        onClick={this.handleMouseClick}
-        onMouseEnter={this.handleMouseEnter}
-        onMouseLeave={this.handleMouseLeave}
-        onMouseOver={this.handleMouseOver}
+        onClick={(e) => this.handleMouseClick(e)}
+        onMouseEnter={this.props.isPopover ? null : this.handleMouseEnter}
+        onMouseLeave={this.props.isPopover ? null : this.handleMouseLeave}
+        onMouseOver={this.props.isPopover ? null : this.handleMouseOver}
```
maciek15: Below you can use Portal component from `@talixo/portal`

packages/tooltip/stories.js
```diff
@@ -109,4 +109,32 @@ addStory('custom fade time', readme, () => (
...
+    </Tooltip>
+  }>
+    <span
+      style={tooltipStyle}
+    >
+      Popover
+    </span>
+  </Tooltip>
+))
+
```
maciek15: I would suggest you can group top, right, left and bottom tooltip in one story, the same goes with fade - you can group it in one story

packages/tooltip/src/Tooltip.js
```diff
@@ -117,24 +107,34 @@ class Tooltip extends React.Component {
...
   handleMouseOver () {
     if (!_.isUndefined(this.props.isOpen)) return
     if (!this.state.clicked && !this.state.isOpen) this.setState({ isOpen: true })
+    this.updatePosition()
   }
 
-  handleMouseClick () {
-    if (!_.isUndefined(this.props.isOpen)) return
-    this.setState({ clicked: true, isOpen: false })
+  handleMouseClick (e) {
```
stuniel: `this.updatePosition()` should be called regardless of `this.state.isOpen` state. This should be refactored, i.e:
```
handleMouseClick () {
  if (!this.props.isPopover) return
  else {
    this.setState({ isOpen: !this.state.isOpen })
    this.updatePosition()
  }
}
```

packages/tooltip/src/Tooltip.js
```diff
@@ -64,23 +66,11 @@ class Portal extends React.Component {
...
-  * @property {function} [props.render]
-  * @property {object} [props.style]
-  */
   constructor (props) {
     super(props)
 
     this.state = {
+      isPopover: this.props.isPopover,
       clicked: false,
       isOpen: this.props.isOpen,
```
stuniel: Our props unification implies that all prop names should remain as rudimentary as possible. I would suggest changing `isOpen` to `open`.

packages/tooltip/src/Tooltip.js
```diff
@@ -64,23 +66,11 @@ class Portal extends React.Component {
...
-  * @property {string} [props.color]
-  * @property {number} [props.fadeTime]
-  * @property {function} [props.render]
-  * @property {object} [props.style]
-  */
   constructor (props) {
     super(props)
 
     this.state = {
+      isPopover: this.props.isPopover,
```
stuniel: I suggest changing this prop to `controlType`, so it could include all behaviours - on click, hover and controlled.

packages/tooltip/styles/geometry.sass
```diff
@@ -63,7 +86,14 @@
       top: 0
       left: 50%
       border-width: 0 dp(0.5) dp(1.25) dp(0.5)
-      border-color: transparent transparent $color-orange-talixo transparent
+      border-color: transparent transparent $color-tab-border transparent
```
rangoo94: We should extract such variables to `config.scss` in component, because why `$color-tab-border` should be in popover?

packages/tooltip/styles/geometry.sass
```diff
@@ -73,21 +103,24 @@
...
+    display: inline-block
     position: relative
     width: auto
     height: auto
+
+  &--arrow
+    &:before
+      display: inline-block
+    &:after
+      display: inline-block
```
rangoo94: Missing empty line on EOF

packages/tooltip/styles/geometry.sass
```diff
@@ -73,21 +103,24 @@
...
 
-  &-hover
+  &__wrapper
+    display: inline-block
     position: relative
     width: auto
     height: auto
+
+  &--arrow
+    &:before
```
rangoo94:
```sass
&--arrow
  &:before,
  &:after
    display: inline-block

File: packages/tooltip/tests/position.test.js [hunk 9]
```diff
@@ -1,6 +1,8 @@
 import { getPositionNearElement } from '../utils/position'
 
 describe('getPositionNearElement', () => {
+  let getBoundingClientRect = window.Element.prototype.getBoundingClientRect
+
   beforeEach(() => {
```
rangoo94: It should be used in `Tooltip.test.js` as well

rangoo94:
@maciek15, @stuniel, could you review it?

packages/tooltip/README.md
```diff
@@ -32,10 +32,13 @@ position      | PositionType | `right` | Tooltip position
 rootNode      | string       | `body`  | Root element of tooltip portal
```
maciek15: `rootNode` is never used, now it is called `attachTo`

packages/tooltip/src/Tooltip.js
```diff
@@ -190,12 +187,20 @@ Tooltip.propTypes = {
...
 
   /** Additional styles passed to the tooltip */
-  style: PropTypes.object
+  style: PropTypes.object,
+
+  /* Show arrow next to tolltip */
+  arrow: PropTypes.bool,
+
+  /** Type of event to open tooltip  */
+  triggerOn: PropTypes.oneOf([ 'hover', 'click' ])
```
maciek15: `propTypes` and `defaultProps` should be before class

packages/tooltip/stories.js
```diff
@@ -56,57 +56,126 @@ function getInitialState () {
...
+      Tooltip bottom
+      </span>
+    </Tooltip>
+  </div>
 ))
-addStory('top', readme, () => (
-  <Tooltip color='primary' position='top' render={() => <span>Default</span>}>
+
+addStory('without arrow', readme, () => (
+  <Tooltip isArrow={false} color='primary' render={() => <div><h3>Without arrow</h3></div>}>
```
maciek15: Could you change to `arrow`? :)

packages/tooltip/src/Tooltip.js
```diff
@@ -80,31 +68,37 @@ class Tooltip extends React.Component {
...
+    this.updatePosition()
   }
 
   handleMouseOver () {
     if (!_.isUndefined(this.props.open)) return
     if (!this.state.clicked && !this.state.open) this.setState({ open: true })
+    this.updatePosition()
   }
 
   handleMouseClick () {
```
maciek15: @rangoo94, should all of these functions (except lifecycle methods) be documented?

packages/tooltip/stories.js
```diff
@@ -17,18 +18,150 @@ const tooltipStyle = {
...
+      render={() => <div><h3>Popover fade 3s</h3></div>}
+    >
+      <span style={tooltipStyle}>
+        Popover
+      </span>
+    </Tooltip>
+  </div>
+))
+
+addStory('Popover with tooltip', readme, () => (
```
maciek15: Position of tooltip in this example should be changed. Right now it looks like this: 
image

PR #58, opened 2018-04-12
[UI-13] Add unit tests for typography

rangoo94:


packages/typography/tests/Heading.test.js
```diff
@@ -0,0 +1,42 @@
...
+        Primary information
+      </Heading>
+    )
+
+    expect(wrapper).toMatchSnapshot()
+  })
+
+  it('should not allow bigger level than 6', () => {
+    const wrapper = shallow(
+      <Heading level={8}>
```
michalochman: Do we want to silently pass this or would it be better to validate max value with props?

rangoo94: Good point, I will change it this way :+1:

PR #71, opened 2018-04-19
[UI-46] Add component for Rating representation

rangoo94:
#### Task

- **Issue or task referred:** UI-46
- **Feature or Bugfix:** Feature

Add component for Rating representation, we will need it soon.

#### Checklist

- [x] Required unit tests prepared
- [x] Documentation prepared
- [x] External licenses validated (ideally MIT)

PR #72, opened 2018-04-19
Bugfix/combo box tests

rangoo94:
#### Task

There was a problem with `combo-box` unit tests, as Downshift is using `setTimeout` after click.


rangoo94:
It is already in `master` through #71 .

PR #75, opened 2018-04-20
[WIP] [UI-26] [UI-27] Handle suffixes and icons in inputs

rangoo94:
#### Task

- **Issue or task referred:** UI-26, UI-27

Make possibility to add icons (or anything on left/right side of input) and suffix (just after value) in the text and number inputs.

#### Checklist

- [ ] Handle suffixes in the text input
- [ ] Handle icons in the text input
- [ ] Use TextInput in NumberInput
- [ ] Handle RTL as well
- [ ] Documentation prepared


rangoo94:
Thanks for that, I just fixed that in stories 👍

michalochman:
LGTM.

2 remarks:
1. You can't change value in "default" story
2. In controlled story, it would be nice to provide `min/max` values (e.g. `0/8`) – it will work nicely in this context.

rangoo94:
The problem with not changing the value in 'default' story we have everywhere; we should think about the generic solution for them (or using always controlled stories?)

packages/combo-box/styles/geometry.sass
```diff
@@ -9,7 +9,7 @@
     width: 100%
     cursor: pointer
 
-  &__button, &__input
+  &__button, &__input input
```
stuniel: This breaks `open` status of AutocompleteInput - `open` modifier is added to `input` instead of `&__input`.

rangoo94: It is changed in `TextInput`

PR #80, opened 2018-04-24
[UI-39] Inline input

stuniel:
#### Task

- **Issue or task referred:** UI-39

A component that represents an inline edit input.

#### Checklist

- [x] Documentation prepared


packages/inline-input/src/InlineInput.js
```diff
@@ -0,0 +1,186 @@
...
+
+    this.setState({ editing, selected, selection })
+  }
+
+  setRef = (node) => {
+    this._input = node
+  }
+
+  render () {
+    const { className, disabled, emptyValue, error, icon, placeholder, value, ...passedProps } = this.props
```
maciek15: There a warning in a console 
> Warning: Unknown event handler property `onInputChange`. It will be ignored.

It can be resolved by spreading onInputChange here so it will not be passed to a div

Additionally, could you split this into lines? :)

packages/inline-input/styles/geometry.sass
```diff
@@ -0,0 +1,25 @@
...
+  position: relative
+  color: $inline-input-text-color
+
+  &__input input
+    width: 100%
+    border: 1px solid $inline-input-border-color
+    border-radius: $inline-input-border-radius
+    background-color: $inline-input-background-color
+    color: $inline-input-text-color
+
```
maciek15: Colors should be in theming.sass

packages/inline-input/src/InlineInput.js
```diff
@@ -0,0 +1,186 @@
...
+      <div className={wrapperClsName} {...passedProps}>
+        {editing && !disabled
+          ? <TextInput
+            className={inputClsName}
+            error={error}
+            inputRef={node => this.setRef(node)}
+            placeholder={placeholder || inputValue}
+            value={inputValue}
+            onChange={(value) => this.handleInputChange(value)}
+            onKeyPress={(e) => this.handleKeyPress(e)}
```
maciek15: `onChange` and `onKeyPress` can be just
`onKeyPress={this.handleKeyPress}`
`onChange={this.handleInputChange}`

or without `this` if you spread them before return (same with onBlur)

packages/inline-input/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Supported props
+
+It allows any properties which are allowed in `div`. Additionally, it handles some differently:
+
+Property name   | Required | Type      | Default | Description
+----------------|----------|-----------|:-------:|--------------------------------
+className       | no       | string    | n/a     | Additional class name passed to wrapper.
+disabled        | no       | boolean   | `false` | Prevents editing the content.
+emptyValue | no       | string    | n/a     | Rendered value for empty input value.
```
rangoo94: Could you fix spaces here?

packages/inline-input/src/InlineInput.js
```diff
@@ -0,0 +1,186 @@
...
+    }
+    obj.focus()
+  }
+
+  handleBlur = () => {
+    this.setState({ editing: false })
+  }
+
+  handleInputChange = (inputValue) => {
+    const { onInputChange } = this.props
```
rangoo94: It should be `onChange`

packages/inline-input/src/InlineInput.js
```diff
@@ -0,0 +1,186 @@
...
+
+  componentDidUpdate (prevProps, prevState) {
+    const editable = this._input !== undefined && prevState.editing !== this.state.editing && !prevState.editing
+
+    if (editable) {
+      this.focusInput(this._input)
+    }
+  }
+
+  focusInput = (obj) => {
```
rangoo94: It don't have to be created each time a component is built: `focusInput (obj) {`

packages/inline-input/styles/geometry.sass
rangoo94: No, colors shouldn't be in theming. Theming should contain stuff which is not required to make it working correctly (i.e. animations, optional styling)

rangoo94: I think that it will be good to remove concept of **geometry & theming** from our project yet :)

packages/inline-input/src/InlineInput.js
rangoo94: and `inputRef` ;)

packages/inline-input/src/InlineInput.js
stuniel: Could you elaborate on this?

packages/inline-input/src/InlineInput.js
```diff
@@ -0,0 +1,186 @@
...
+    const { selection, selected } = this.state
+    const { start, end } = selection
```
maciek15: As selection is never used this one could be spread on top like this:
`const { selected, selection: { start, end } } = this.props`

packages/inline-input/stories.js
```diff
@@ -0,0 +1,71 @@
...
+    />
+    <span style={{ display: 'inline-block', padding: '16px', fontWeight: 700 }}>
+      {/* You typed: {state.value} */}
+    </span>
+  </div>
+), () => {
+  return {
+    value: ''
+  }
+})
```
maciek15: I seems that using storybook action addon would be enough in this case

`const change = action('change')`

packages/inline-input/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@talixo/inline-input",
+  "version": "1.0.0",
```
rangoo94: `0.0.0`

packages/inline-input/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "devDependencies": {
+    "@talixo/icon": "^1.0.0"
+  },
+  "peerDependencies": {
+    "prop-types": "^15.6.1",
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0"
+  },
+  "dependencies": {
+    "@talixo/shared": "^1.0.0",
```
rangoo94: Move `dependencies` to both `devDependencies` and `peerDependencies`, add information to README file

packages/inline-input/src/InlineInput.js
```diff
@@ -0,0 +1,224 @@
...
+   *
+   */
+  handleKeyPress = (e) => {
+    if (e.key === 'Enter') { this.setState({ editing: false }) }
+  }
+
+  /**
+   * Handle span click.
+   *
+   * @param {SyntheticEvent} e   *
```
rangoo94: What is that `*` on end?

packages/inline-input/src/InlineInput.js
```diff
@@ -0,0 +1,224 @@
...
+    this.setState({ inputValue })
+
+    // Trigger input change to parent components
+    if (onChange) { onChange(inputValue) }
+  }
+
+  /**
+   * Handle key press.
+   *
+   * @param {SyntheticEvent} e   *
```
rangoo94: What is that `*` on end?

packages/inline-input/src/InlineInput.js
```diff
@@ -0,0 +1,224 @@
...
+      selection = defaultSelection
+    }
+
+    this.setState({ editing, selected, selection })
+  }
+
+  /**
+   * Set ref.
+   *
+   * @param {node} node   *
```
rangoo94: What is that `*` on end? Also, in JS there is no `node` type

PR #99, opened 2018-05-08
Feature/UI 45 Wizard component

PiotrGrzechnik:
#### Task

Wizard component display each step, one by one. It can be used for example to multi-step form.
As children It can accept other component or DOM element. Please note that each child is treat as one step in order: first child is step one, second child is step two, etc...

#### Checklist

- [ ] Documentation prepared


packages/wizard/src/Wizard.js
```diff
@@ -0,0 +1,66 @@
...
+ * @property {string} [props.nextLabel]
+ * @property {string} [props.previousLabel]
+ * @class {React.Element}
+ */
+class Wizard extends React.PureComponent {
+  constructor (props) {
+    super(props)
+    this.state = {
+      currentStep: 1
+    }
```
maciek15: It should also handle passing current step from props. Here it would be like this:
`
this.state = {
  currentStep: props.step || 1
 }
`
And additionally it should check if this prop have been change in `componentWillReceiveProps`

PR #100, opened 2018-05-11
[UI-17] Investigate and set up examples for forms

maciek15:
#### Task

- **Issue or task referred:** UI-17

I have added Formik wrapper component created by @rangoo94 along with some modifications and documentation. Additionally I have created some examples of how to use forms to storybook.

#### Checklist

- [ ] Documentation prepared


packages/form-handler/styles/geometry.sass
```diff
@@ -0,0 +1,4 @@
+@import './config'
+
+@include module('form-handler')
+  display: block
```
PiotrGrzechnik: I think It's not necessary here, because form is block by default.

packages/form-handler/styles/theming.sass
```diff
@@ -0,0 +1,4 @@
+@import './config'
+
+@include module('form-handler')
+  display: block
```
PiotrGrzechnik: The same as above. Not necessary line.

PR #105, opened 2018-05-18
Feature/UI-56 Slider component

PiotrGrzechnik:
#### Task

Slider component lets you choose value from initial (0-100) or custom range (props: `min` & `max`).
It can also has `label` passed as props. You can set `step` as size of each movement of the slider control. Additionally you can set `defaultValue`.

#### Checklist

- [x] Documentation prepared

### DIFF
--- a/packages/slider/src/Slider.js
+++ b/packages/slider/src/Slider.js
@@ -9,6 +9,12 @@ const propTypes = {
   /** Additional class name */
   className: PropTypes.string,
 
+  /** Defaul value */
+  defaultValue: PropTypes.number,
+
+  /** Label for input range */
+  label: PropTypes.string,
+
   /** Maximum value in range */
   max: PropTypes.number,
 
@@ -24,6 +30,8 @@ const propTypes = {
  *
  * @property {object} props
  * @property {string} [props.className]
+ * @property {number} [props.defaultValue]
+ * @property {string} [props.label]
  * @property {number} [props.max]
  * @property {number} [props.min]
  * @property {number} [props.step]
--- a/packages/slider/src/Slider.js
+++ b/packages/slider/src/Slider.js
@@ -1,0 +1,90 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { buildClassName } from '@talixo/shared'
+
+const moduleName = 'slider'
+
+const propTypes = {
+  /** Additional class name */
+  className: PropTypes.string,
+
+  /** Maximum value in range */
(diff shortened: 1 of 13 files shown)


packages/slider/src/Slider.js
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * This function handle change of input value
+   * @param {object} e
+   */
+  handleChange (e) {
+    this.setState({
+      value: e.target.value
+    })
+  };
```
maciek15: I does not handle external onChange event properly. As our other components it should pass value to this.props.onChange.

packages/slider/src/Slider.js
```diff
@@ -0,0 +1,98 @@
...
+
+  render () {
+    const { className, label, max, min, step, ...passedProps } = this.props
```
maciek15: You should also add onChnge, onBlur and onFocus here so it will not be passed to span.

packages/slider/src/Slider.js
```diff
@@ -0,0 +1,98 @@
...
+    return (
+      <span className={buildClassName(moduleName, className)} {...passedProps}>
+        {label && <label className={buildClassName([moduleName, 'label'])}>{label}</label>}
+        <span className={buildClassName([moduleName, 'value'])}>{value}</span>
+        <input
+          className={buildClassName([moduleName, 'input'])}
+          max={max}
+          min={min}
+          onChange={this.handleChange}
+          step={step}
```
maciek15: You should also handle onFocus and onBlur events.

packages/slider/stories.js
```diff
@@ -0,0 +1,34 @@
...
+
+// Create factories for story
+const addStory = createStoriesFactory('Slider', module, {
+  propTables: [ Slider ]
+})
+
+// Stories
+
+addStory('initial', readme, () => (
+  <Slider />
```
maciek15: After you will add external onChange, onBlur and onFocus handlers, you may also add actions for these events from `@storybook/addon-actions`

packages/slider/styles/geometry.sass
```diff
@@ -0,0 +1,31 @@
...
+  &__input
+    -webkit-appearance: none
+    width: 100%
+    height: dp(0.275)
+    cursor: pointer
+    transition: background .2s linear
+    outline: none
+
+    &::-webkit-slider-thumb
+      -webkit-appearance: none;
```
maciek15: This semicolon can be removed

PR #110, opened 2018-05-30
[UI-59] Split view

maciek15:
#### Task

- **Issue or task referred:** UI-59

i have created component which show a list of customizable data items and its details which can be either self- or externally controlled.

#### Checklist

- [x] Documentation prepared


packages/split-view/styles/config.sass
```diff
@@ -0,0 +1,13 @@
+@import '~@talixo/shared/styles/config'
+
+// Dimensions
+$split-view-list-width: dp(50) !default
+$split-view-border: solid 1px $color-border !default
```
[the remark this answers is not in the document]
rangoo94: Actually, he could use `$border-width` instead, but let keep it after `feature/new-colors`

packages/split-view/src/DetailsView.js
```diff
@@ -0,0 +1,52 @@
...
+ * Component which renders details about given data item
+ * using provided detailsRender function.
+ *
+ * @param {object} props
+ * @param {function} [props.datailsRender]
+ * @param {object} [props.item]
+ *
+ * @returns {React.Element}
+ */
+const DetailsView = (props) => {
```
rangoo94: We use `function` instead for components

packages/split-view/src/DetailsView.js
```diff
@@ -0,0 +1,52 @@
...
+  const { detailsRender, item } = props
+  const detailsCls = buildClassName([moduleName, 'details'])
+
+  return (
+    <React.Fragment>
```
rangoo94: Why it's a `React.Fragment`? Shouldn't

```jsx
function DetailsView (props) {
  const { detailsRender, item } = props

  const detailsCls = buildClassName([ moduleName, 'details' ])

  return item != null ? (
    <div className={detailsCls}>
      {detailsRender(item)}
    </div>
  )
}
```

packages/split-view/src/DetailsView.js
```diff
@@ -0,0 +1,52 @@
...
+      {
+        item
+          ? <div className={detailsCls}>
+            { item && detailsRender(item) }
```
rangoo94: In this place you have `item` validated already

packages/split-view/src/DetailsView.js
```diff
@@ -0,0 +1,52 @@
...
+import PropTypes from 'prop-types'
+
+import { buildClassName } from '@talixo/shared'
+
+import { moduleName } from './config'
+import { displayObject } from './utils'
+
+const propTypes = {
+  /** Function used to render data details. */
+  detailsRender: PropTypes.func,
```
rangoo94: Wouldn't be better to change it to `renderDetails`?

packages/split-view/src/Item.js
```diff
@@ -0,0 +1,57 @@
...
+import PropTypes from 'prop-types'
+
+import { buildClassName } from '@talixo/shared'
+
+import { moduleName } from './config'
+import { displayObject } from './utils'
+
+const propTypes = {
+  /** Function used to render data item. */
+  itemRender: PropTypes.func,
```
rangoo94: Wouldn't be better to change it to `renderItem`?

packages/split-view/src/Item.js
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param {object} props
+ * @param {bool} [props.active]
+ * @param {object} props.item
+ * @param {function} [props.itemRender]
+ * @param {function} [props.onClick]
+ *
+ * @returns {React.Element}
+ */
+const Item = (props) => {
```
rangoo94: We use `function` for components

packages/split-view/src/Item.js
```diff
@@ -0,0 +1,57 @@
...
+  const { active, item, itemRender, onClick } = props
+  const itemCls = buildClassName([moduleName, 'list', 'item'], null, { active })
```
rangoo94: Avoid nested elements (`[ moduleName, 'list-item' ]`)

packages/split-view/src/ItemList.js
```diff
@@ -0,0 +1,63 @@
...
+
+import { moduleName } from './config'
+
+const propTypes = {
+
+  /** Additional class name. */
+  className: PropTypes.string,
+
+  /** Function used to render data item. */
+  itemRender: PropTypes.func,
```
rangoo94: `renderItem`

packages/split-view/src/ItemList.js
```diff
@@ -0,0 +1,63 @@
...
+
+  /** Array of objects which contain data to be displayed. */
+  items: PropTypes.arrayOf(PropTypes.object).isRequired,
+
+  /** onClick callback function. */
+  onClick: PropTypes.func,
+
+  /** Currently opened item. */
+  openItem: PropTypes.object
```
rangoo94: According to description it's actually `openedItem` or `opened`, or `value` to keep it consistent ;)

packages/split-view/src/SplitView.js
```diff
@@ -0,0 +1,124 @@
...
+
+const propTypes = {
+  /** Additional class name. */
+  className: PropTypes.string,
+
+  /** Data to be displayed inside SplitView component. */
+  data: PropTypes.arrayOf(PropTypes.object).isRequired,
+
+  /** Render function of details section. */
+  detailsRender: PropTypes.func,
```
rangoo94: `renderDetails`

packages/split-view/src/SplitView.js
```diff
@@ -0,0 +1,124 @@
...
+
+  /** Render function of list section. */
+  itemRender: PropTypes.func,
```
rangoo94: `renderItem`

packages/split-view/src/SplitView.js
```diff
@@ -0,0 +1,124 @@
...
+
+  /** Header element of list section. */
+  listHeader: PropTypes.node,
+
+  /** onSelect callback. */
+  onSelect: PropTypes.func,
```
rangoo94: **If** opened item will be `value` **then** it could be `onChange`

packages/split-view/src/ItemList.js
rangoo94: Is `openItem` an `object` or `number`? I am asking, because in `change` method there is check for `openItem === itemIndex`

packages/split-view/src/SplitView.js
maciek15: @michalochman, what do you think about naming this one?

packages/split-view/src/ItemList.js
maciek15: There's no such thing as `openItem === itemIndex` inside code :smile:

maciek15: That was inside `SplitView`, thanks :smile:

packages/split-view/src/SplitView.js
michalochman: @maciek15 do what @rangoo94 is suggesting, we have already merged latest changes which introduce this naming convention

packages/split-view/src/Item.js
```diff
@@ -0,0 +1,57 @@
...
+  renderItems: PropTypes.func,
+
+  /** Object which contains data to be displayed. */
+  item: PropTypes.object.isRequired,
+
+  /** onClick callback function. */
+  onClick: PropTypes.func
+}
+const defaultProps = {
+  active: false,
```
michalochman: This prop is not defined in `propTypes`. The other thing is that it would not have to be defaulted to `false`, it would be falsy if not required and not provided.

packages/split-view/src/SplitView.js
```diff
@@ -0,0 +1,122 @@
...
+
+    if (onSelect) { onSelect(item, ...args) }
+    if (propsOpenedItem == null && openedItem !== item) {
+      this.setState({ openedItem: item })
+    }
+  }
+  /**
+   * Generates header component if provided.
+   */
+  generateHeader = () => {
```
michalochman: Maybe `renderHeader`?

PR #113, opened 2018-06-08
[UI-64] Chat

stuniel:
#### Task

- **Issue or task referred:** UI-64

We prepared a Chat component which can be used as a chat or a comments section.  It uses renderers to add additional information, buttons and a render prop to provide message renderer. The compononent accepts an array of messages and returns a message on `onSubmit`. It additionally handles user typing via `addTypingUser` and `usersTyping` props.

#### Checklist

- [ ] Documentation prepared

### DIFF
--- a/packages/chat/README.md
+++ b/packages/chat/README.md
@@ -32,6 +32,7 @@ className             | string        | n/a                  | Additional class 
 messages              | Messages[]    | `[]`                 | Array of messages.
 messageRenderer       | func          | `message => message` | Message renderer.
 name                  | string        | `'user'`             | User name.
+onSubmit              | func          | n/a                  | Handler for onSubmit event.
 id                    | string        | Required             | User id.
 usersTyping           | UsersTyping[] | `[]`                 | Typing users.
 placeholder           | string        | `'reply'`            | Reply input placeholder.
--- a/packages/chat/README.md
+++ b/packages/chat/README.md
@@ -1,0 +1,30 @@
+# Talixo Chat
+
+UI Component which represents Chat
+
+## How to install
+
+Package is available as `@talixo/chat` in NPM registry, so you can use it in your project
+using `npm install @talixo/chat --save` or `yarn add @talixo/chat`.
+
+## Requirements
+
+Your package should additionally have some extra dependencies:
+
+- `prop-types: ^15.6.1`
+- `react: ^16.2.0`
+- `react-dom: ^16.2.0`
+- `@talixo/shared: ^0.1.0`
+
+These packages are required by `@talixo/chat`, but you have to install them manually,
+to avoid having different versions of these in your application.
+
+## Supported props
+
+Property name | Type      | Default | Description                    
+--------------|-----------|:-------:|--------------------------------
+className     | string    | n/a     | additional class name passed to wrapper
+
+## Changelog
+
+- **0.1.0** - initial version
--- a/packages/chat/README.md
+++ b/packages/chat/README.md
@@ -29,9 +29,10 @@ Property name         | Type          | Default              | Description
 additionalButton      | node          | n/a                  | Additional button.
 additionalInformation | node          | n/a                  | Information message.
 className             | string        | n/a                  | Additional class name.
-messages              | Messages[]    | `[]`                 | Additional class name.
+messages              | Messages[]    | `[]`                 | Array of messages.
 messageRenderer       | func          | `message => message` | Message renderer.
-user                  | string        | `'user'`             | User name.
+name                  | string        | `'user'`             | User name.
+id                    | string        | Required             | User id.
 usersTyping           | UsersTyping[] | `[]`                 | Typing users.
 type                  | Type          | `'chat'`             | Message type.
 
@@ -47,7 +48,8 @@ Type      | 'chat', 'comments'
 
 Property name | Type      | Required | Default       | Description
 --------------|-----------|----------|:-------------:|------------------------------------------------
-user          | string    | no       | n/a           | User name.
+name          | string    | no       | n/a           | User name.
+id            | string    | no       | n/a           | User id.
 message       | node      | no       | n/a           | Message content.
 time          | number    | no       | n/a           | Message time stamp.
 
@@ -55,7 +57,8 @@ time          | number    | no       | n/a           | Message time stamp.
 
 Property name | Type      | Required | Default       | Description
 --------------|-----------|----------|:-------------:|------------------------------------------------
-user          | string    | no       | n/a           | User name.
+name          | string    | no       | n/a           | User name.
+id            | string    | no       | n/a           | User id.
 status        | boolean   | no       | n/a           | Typing status.
 
 ## Changelog
--- a/packages/chat/README.md
+++ b/packages/chat/README.md
@@ -15,6 +15,7 @@ Your package should additionally have some extra dependencies:
 - `react: ^16.2.0`
 - `react-dom: ^16.2.0`
 - `@talixo/shared: ^0.1.0`
+- `@talixo/icon: ^0.2.0`
 
 These packages are required by `@talixo/chat`, but you have to install them manually,
 to avoid having different versions of these in your application.
--- a/packages/chat/README.md
+++ b/packages/chat/README.md
@@ -22,10 +22,42 @@ to avoid having different versions of these in your application.
 
 ## Supported props
 
-Property name | Type      | Default | Description                    
---------------|-----------|:-------:|--------------------------------
-className     | string    | n/a     | additional class name passed to wrapper
+It allows any props which are allowed for `div`. Additionally, it handles some differently:
+
+Property name         | Type          | Default              | Description
+----------------------|---------------|:--------------------:|--------------------------------
+additionalButton      | node          | n/a                  | Additional button.
+additionalInformation | node          | n/a                  | Information message.
+className             | string        | n/a                  | Additional class name.
+messages              | Messages[]    | `[]`                 | Additional class name.
+messageRenderer       | func          | `message => message` | Message renderer.
+user                  | string        | `'user'`             | User name.
+usersTyping           | UsersTyping[] | `[]`                 | Typing users.
+type                  | Type          | `'chat'`             | Message type.
+
+#### Types
+
+Type name | Enum options
+----------|---------------------------------------------------
+Type      | 'chat', 'comments'
+
+### Property shapes
+
+#### Messages
+
+Property name | Type      | Required | Default       | Description
+--------------|-----------|----------|:-------------:|------------------------------------------------
+user          | string    | no       | n/a           | User name.
+message       | node      | no       | n/a           | Message content.
+time          | number    | no       | n/a           | Message time stamp.
+
+#### UsersTyping
+
+Property name | Type      | Required | Default       | Description
+--------------|-----------|----------|:-------------:|------------------------------------------------
+user          | string    | no       | n/a           | User name.
+status        | boolean   | no       | n/a           | Typing status.
 
 ## Changelog
 
-- **0.1.0** - initial version
+- **0.0.0** - initial version
--- a/packages/chat/README.md
+++ b/packages/chat/README.md
@@ -34,6 +34,7 @@ messageRenderer       | func          | `message => message` | Message renderer.
 name                  | string        | `'user'`             | User name.
 id                    | string        | Required             | User id.
 usersTyping           | UsersTyping[] | `[]`                 | Typing users.
+placeholder           | string        | `'reply'`            | Reply input placeholder.
 type                  | Type          | `'chat'`             | Message type.
 
 #### Types
--- a/packages/chat/README.md
+++ b/packages/chat/README.md
@@ -27,6 +27,7 @@ It allows any props which are allowed for `div`. Additionally, it handles some d
 Property name         | Type          | Default              | Description
 ----------------------|---------------|:--------------------:|--------------------------------
 additionalButton      | node          | n/a                  | Additional button.
+addTypingUser         | func          | n/a                  | Called when the user is typing.
 additionalInformation | node          | n/a                  | Information message.
 className             | string        | n/a                  | Additional class name.
 messages              | Messages[]    | `[]`                 | Array of messages.
--- a/packages/chat/src/Chat.js
+++ b/packages/chat/src/Chat.js
@@ -23,7 +23,10 @@ const propTypes = {
   messages: PropTypes.arrayOf(
     PropTypes.shape({
       /** User name. */
-      user: PropTypes.string,
+      name: PropTypes.string,
+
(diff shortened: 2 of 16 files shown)


packages/chat/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+It allows any props which are allowed for `div`. Additionally, it handles some differently:
+
+Property name         | Type          | Default              | Description
+----------------------|---------------|:--------------------:|--------------------------------
+additionalButton      | node          | n/a                  | Additional button.
+addTypingUser         | func          | n/a                  | Called when the user is typing.
+additionalInformation | node          | n/a                  | Information message.
+className             | string        | n/a                  | Additional class name.
+messages              | Messages[]    | `[]`                 | Array of messages.
```
rangoo94: `Message[]`

packages/chat/README.md
```diff
@@ -0,0 +1,71 @@
...
+messageRenderer       | func          | `message => message` | Message renderer.
+name                  | string        | `'user'`             | User name.
+onSubmit              | func          | n/a                  | Handler for onSubmit event.
+id                    | string        | Required             | User id.
+usersTyping           | UsersTyping[] | `[]`                 | Typing users.
```
rangoo94: Shouldn't it be just a list of names to show?

rangoo94: Also, it's more likely `TypingUser[]`

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+ * @property {string} [props.messages.name]
+ * @property {string} [props.messages.id]
+ * @property {*} [props.messages.message]
+ * @property {number} [props.messages.time]
+ * @property {*} [props.messageRenderer]
+ * @property {string} [props.name]
+ * @property {*} [props.onSubmit]
+ * @property {string} [props.id]
+ * @property {array} [props.usersTyping]
+ * @property {string} [props.usersTyping.name]
```
rangoo94: It is not true - `props.usersTyping` is an array, so it doesn't have `name`, `id` or `status` property. You should either create `@type`, or use `@property {Array<{ [name]: string, [id]: string, [status]: boolean }>} [props.usersTyping]`.

Also, really all elements are optional?

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+ * @property {string} [props.className]
+ * @property {array} [props.messages]
+ * @property {string} [props.messages.name]
+ * @property {string} [props.messages.id]
+ * @property {*} [props.messages.message]
+ * @property {number} [props.messages.time]
+ * @property {*} [props.messageRenderer]
+ * @property {string} [props.name]
+ * @property {*} [props.onSubmit]
+ * @property {string} [props.id]
```
rangoo94: What is that `id` for?

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+ * @property {*} [props.additionalInformation]
+ * @property {*} [props.addTypingUser]
+ * @property {string} [props.className]
+ * @property {array} [props.messages]
+ * @property {string} [props.messages.name]
+ * @property {string} [props.messages.id]
+ * @property {*} [props.messages.message]
+ * @property {number} [props.messages.time]
+ * @property {*} [props.messageRenderer]
+ * @property {string} [props.name]
```
rangoo94: What is that `name` for

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+/**
+ * Component which represents Chat.
+ *
+ * @property {object} props
+ * @property {*} [props.additionalButton]
+ * @property {*} [props.additionalInformation]
+ * @property {*} [props.addTypingUser]
+ * @property {string} [props.className]
+ * @property {array} [props.messages]
+ * @property {string} [props.messages.name]
```
rangoo94: Same problem as with `usersTyping`

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+  name: 'user'
+}
+
+/**
+ * Component which represents Chat.
+ *
+ * @property {object} props
+ * @property {*} [props.additionalButton]
+ * @property {*} [props.additionalInformation]
+ * @property {*} [props.addTypingUser]
```
rangoo94: `onTyping`? There is only one user who can start typing in context of this chat

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+    typingStatus: false
+  }
+
+  /**
+   * Fire addTypingUser when state.typingStatus is updated.
+   * Scroll messages continer to bottom if messages are updated.
+   */
+  componentDidUpdate (prevProps, prevState) {
+    if (prevState.typingStatus !== this.state.typingStatus && this.props.addTypingUser) {
+      this.props.addTypingUser({
```
rangoo94: From this context we should only said that user is typing

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+        status: this.state.typingStatus,
+        name: this.props.name,
+        id: this.props.id
+      })
+    }
+
+    if (this.props.messages !== prevProps.messages) {
+      this.scrollToBottom(this._messages)
```
rangoo94: Probably there should be an option to scroll to bottom only when it's currently on the bottom.

Imagine a situation when the user is spammed, so he can't go back to the top of messages.

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+   * Handle input value change.
+   *
+   * @param {string} inputValue
+   */
+  handleInputChange = (e) => {
+    const inputValue = e.target.value
+
+    // If the user is typing clear the timeout
+    if (this.state.typingStatus) {
+      clearTimeout(this._typingTimeout)
```
rangoo94: It should be cleared on `unmount` as well

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+    // Build message
+    const message = {
+      message: inputValue,
+      name: name,
+      id: id,
+      time: moment().valueOf()
+    }
+
+    // Submit message if the inputValue is not empty
+    if (onSubmit && !/^(^$|\s+)$/.test(inputValue)) {
```
rangoo94: `/^(^$|\s+)$/` is reduntant, as it's checking twice start and end. More likely `/^\s*$/` is what you had in mind.

rangoo94: Also, it should handle all other stuff even when it doesn't have `onSubmit`. You could also use it as `control statement` to make it more clear:

```js
// ...

if (/^\s*$/.test(inputValue)) {
  return
}

if (onSubmit) {
  onSubmit(message)
}

this.setState({ inputValue: '' })

if (this._input) {
  this._input.style.height = '1em'
}
```

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+      onSubmit(message)
+      this.setState({ inputValue: '' })
+      if (this._input) this._input.style.height = '1em'
```
rangoo94: Why are you setting input height to `1em`? It should be styled within CSS styles directly, or calculated from them. `1em` is magic size here, and I am pretty sure that it's not what actually we want - for example, does it have any padding?

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+    }
+  }
+
+  handleKeyDown = (event) => {
+    const which = event.which || event.key
+    const segments = this._input.value.split('\n')
+    const segmentsLength = segments.length
+
+    let height
+    if (which === 16) {
```
rangoo94: Could you describe what is happening here?

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+  handleKeyUp = (event) => {
+    const which = event.which || event.key
+
+    if (which === 16) {
+      this.setState({ shift: false })
+    }
+
+    if (!this.state.shift && which === 13 && /^(^$|\s+)$/.test(this._input.value)) {
```
rangoo94: Could you describe what is happening here?

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+      this.setState({ inputValue: '' })
+      this._input.style.height = '1em'
+    }
+  }
+
+  handleKeyPress = (event) => {
+    const which = event.which || event.key
+
+    if (!this.state.shift && which === 13) {
```
rangoo94: Could you describe what is happening here?

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+    )
+  }
+
+  /**
+  * Save base node element.
+  *
+  * @param {Element} node
+  * @param {string} name
+  */
+  setRef = (node, name) => {
```
rangoo94: As you've got only to elements to save ref, probably it's better to create two functions instead.

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,356 @@
...
+    const wrapperClsName = buildClassName(moduleName, className)
+    const messagesClsName = buildClassName([moduleName, 'messages'])
+    const formClsName = buildClassName([moduleName, 'form'])
+    const additionalInfoCls = buildClassName([moduleName, 'additional-info'])
+    const additionalBtnCls = buildClassName([moduleName, 'additional-button'])
+    const inputContainerCls = buildClassName([moduleName, 'input-container'])
+    const inputContainerInnerCls = buildClassName([moduleName, 'input-container-inner'])
+
+    return (
+      <div className={wrapperClsName} style={{ display: 'block' }} {...passedProps}>
```
rangoo94: Why it does have `display: block` style passed here?

packages/chat/src/Message.js
```diff
@@ -0,0 +1,58 @@
...
+const defaultProps = {}
+
+function createMarkup (message) {
+  return {__html: message}
+}
+
+function Message (props) {
+  const { className, message, name, time, ...passedProps } = props
+
+  function renderTime () {
```
rangoo94: Could you extract this function outside?

packages/chat/stories.js
```diff
@@ -0,0 +1,268 @@
...
+    time: 1528104696738,
+    message: 'This is message',
+    name: 'John',
+    id: '1'
+  },
+  {
+    time: 1528104730633,
+    message: 'This is reply',
+    name: 'Tom',
+    id: '2'
```
rangoo94: I am not sure if this ID should really be required, probably it should be optional.

packages/chat/stories.js
```diff
@@ -0,0 +1,268 @@
...
+
+const messages = [
+  {
+    time: 1528104696738,
+    message: 'This is message',
+    name: 'John',
+    id: '1'
+  },
+  {
+    time: 1528104730633,
```
rangoo94: Just to make sure, will it allow ISO format as well?

packages/chat/stories.js
```diff
@@ -0,0 +1,268 @@
...
+function stopSimulation (setState, state) {
+  clearInterval(simulation)
+
+  setState({ simulation: false })
+}
+
+// Helpers
+
+const thumbup = {
+  time: moment().valueOf(),
```
rangoo94: Just to note: it's equivalent to `+moment()` (which is calling `valueOf` as well)

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,377 @@
...
+
+  handleKeyDown = (event) => {
+    const which = event.which || event.key
+    const segments = this._input.value.split('\n')
+    const segmentsLength = segments.length
+
+    let height
+    if (which === 16) {
+      this.setState({ shift: true })
+    } else if (this.state.shift && which === 13) {
```
maciek15: It might be better to use object which describes keybord keys used here. Like this or something similar:
`
const keys: { KEY_ENTER: 16, KEY_SHIFT: 13 }
`

packages/chat/src/Message.js
```diff
@@ -0,0 +1,58 @@
...
+  time: PropTypes.number
+}
+
+const defaultProps = {}
+
+function createMarkup (message) {
+  return {__html: message}
+}
+
+function Message (props) {
```
maciek15: Missing JSDoc

packages/chat/src/ReplyBox.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+import { TextInput } from '@talixo/text-input'
+
+function ReplyBox (props) {
```
maciek15: Missing JSDoc

packages/chat/src/ReplyBox.js
```diff
@@ -0,0 +1,14 @@
...
+  return (
+    <TextInput
+      InputComponent='textarea'
+      {...props}
+    />
+  )
+}
+
+export default ReplyBox
```
maciek15: This component is never used

packages/chat/README.md
```diff
@@ -0,0 +1,71 @@
...
+to avoid having different versions of these in your application.
+
+## Supported props
+
+It allows any props which are allowed for `div`. Additionally, it handles some differently:
+
+Property name         | Type          | Default              | Description
+----------------------|---------------|:--------------------:|--------------------------------
+additionalButton      | node          | n/a                  | Additional button.
+addTypingUser         | func          | n/a                  | Called when the user is typing.
```
maciek15: This prop is never used, probably it should be `onTyping` as it is used inside component but not described.

packages/chat/README.md
```diff
@@ -0,0 +1,71 @@
...
+additionalInformation | node          | n/a                  | Information message.
+className             | string        | n/a                  | Additional class name.
+messages              | Message[]     | `[]`                 | Array of messages.
+messageRenderer       | func          | `message => message` | Message renderer.
+name                  | string        | `'user'`             | User name.
```
maciek15: `name` prop should be substituted with `user` and prop shape of user should be added as it really is inside this component. Same thing with `id` prop. It would also be good if you sort those props alphabetically

packages/chat/src/Chat.js
```diff
@@ -0,0 +1,377 @@
...
+ * @property {string} [props.user.id]
+ * @property {string} [props.user.name]
+ * @property {*} [props.onSubmit]
+ * @property {Array<{ [user]: { [name]: string, [id]: string }, [status]: boolean }>} [props.typingUsers]
+ * @property {string} [props.placeholder]
+ * @property {string} [props.type]
+ *
+ * @property {object} state
+ * @property {string} state.inputValue
+ * @property {boolean} state.typingStatus
```
maciek15: Missing `state.shift`

PR #119, opened 2018-06-18
Feature/new colors

rangoo94:
#### Task

- **Issue or task referred:** UI-50

Sort colors and other styles, fix problems with components.

It is not done for all components yet, so we will publish only ready part as stable.

#### Checklist

- [x] Documentation prepared


packages/notification/stories.js
```diff
@@ -2,93 +2,259 @@ import React from 'react'
 import { createStoriesFactory, getReadmeDescription } from '@talixo/shared/story'
 
 import Notification from './src/Notification'
-import NotificationList from './src/NotificationList'
+import NotificationsList from './src/NotificationsList'
```
michalochman: Why have you changed this? "Notification list" is **the** grammatically correct version.

packages/options-input/src/OptionsInputList.js
```diff
@@ -0,0 +1,78 @@
...
+import { buildClassName } from '@talixo/shared'
+
+import ListOption from './ListOption'
+
+const propTypes = {
+  /** Additional class name */
+  className: PropTypes.string,
+
+  /** Options to show */
+  options: PropTypes.arrayOf(PropTypes.shape({
```
michalochman: This could be extracted, it's used more than once.

packages/phone-input/src/PhoneInput.js
```diff
@@ -84,15 +101,23 @@ function renderCountryItem (country) {
...
   state = {
     value: this.props.value || '',
-    focused: false,
-    country: detectCountry(this.props.value)
+    country: detectCountry(this.props.value),
+
+    hover: false,
+    focus: false,
+
+    // Hack for text-mask-input
```
michalochman: Is this long term solution?

packages/phone-input/src/PhoneInput.js
```diff
@@ -143,11 +176,19 @@ class PhoneInput extends React.PureComponent {
...
+    this.change(nextValue)
 
-    // Focus input when it is possible
+    // Focus input when it is possible,
+    // and move cursor after prefix
     if (this.el) {
       this.el.focus()
+
+      if (endPrefix) {
+        this.caretTimeout = setTimeout(() => setCaretPosition(this.el, endPrefix))
```
michalochman: We probably should also clear previously set timeout here (like it's done with `focusTimeout`)

packages/phone-input/src/PhoneInput.js
```diff
@@ -160,6 +201,62 @@ class PhoneInput extends React.PureComponent {
...
+    this.setState({ hover })
+  }
+
+  /**
+   * Set dis-hover state
+   *
+   * @param {string} what
+   */
+  setMouseOutState (what) {
+    setTimeout(() => {
```
michalochman: What is the reason we need this delayed? Looks like we don't need this for `over` state.

packages/phone-input/src/PhoneInput.js
```diff
@@ -160,6 +201,62 @@ class PhoneInput extends React.PureComponent {
...
+    this.setState({ focus })
+  }
+
+  /**
+   * Set blur state
+   *
+   * @param {string} what
+   */
+  setBlurState (what) {
+    setTimeout(() => {
```
michalochman: Is the reason the same as in `out` state?

packages/tooltip/src/Tooltip.js
```diff
@@ -107,6 +112,7 @@ class Tooltip extends React.Component {
   componentDidMount () {
     this.updatePosition()
     window.addEventListener('resize', this.updatePosition)
+    window.addEventListener('scroll', this.updatePosition)
```
michalochman: What do you think about using rAF here for both scroll and resize events? This way scrolling/resizing performance would probably never be degraded if we saved previous values and returned early if we detect there are no changes.

packages/accordion/src/Accordion.js
```diff
@@ -87,6 +130,27 @@ function buildElement (props, option, index) {
...
 }
 
+/**
+ * Check if both lists contain same elements
+ *
+ * @param {array} prevList
+ * @param {array} nextList
+ * @returns {boolean}
+ */
+function isSameList (prevList, nextList) {
```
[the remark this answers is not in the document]
rangoo94: Actually, as `isSameList` is intended for internal use in `Accordion`, in such situation it will work correctly - it will detect that `1` and `2` is selected on both lists.

packages/accordion/src/Accordion.js
```diff
@@ -97,18 +161,19 @@ function buildElement (props, option, index) {
...
  * @property {object} state
- * @property {*} state.value
+ * @property {array} state.value
  *
  * @class
  */
 class Accordion extends React.PureComponent {
   state = {
-    value: this.props.value
+    value: this.props.value == null ? [] : [].concat(this.props.value)
```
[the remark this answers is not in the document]
rangoo94: It is actually not violating single source of truth, it *is* the single point of truth. This way, we are using `value` only from `state`, and `props` is only treat as data-source.

We were using fully controlled components, but it came out that (especially in Accordion situation) setting full control of them within the application is kinda boilerplate.

We could split responsibility, but I am not sure if it's expected - when a developer is using i.e. `input` element he expects that it will work even without handlers; shouldn't we make it in default React behavior?

packages/badge/src/Badge.js
```diff
@@ -12,23 +37,30 @@ import { buildClassName } from '@talixo/shared'
...
+  const { children, className, type, pill, onRemove, removeText, ...passedProps } = props
 
-  const clsName = buildClassName('badge', className)
+  const clsName = buildClassName('badge', className, [ type ], {
+    pill,
+    interactive: !!props.onClick
+  })
+
+  const removeButton = onRemove ? (
+    <span className={buildClassName('badge__remove')} onClick={onRemove}>
```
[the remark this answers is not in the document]
rangoo94: :+1:

packages/button/tests/Button.test.js
```diff
@@ -13,24 +13,24 @@ describe('<Button />', () => {
...
     )
 
     expect(wrapper.props().className).toMatch(/(^| )talixo-button--small( |$)/)
   })
 
-  it('handles `variant` properly', () => {
+  it('handles `wide` properly', () => {
     const wrapper = shallow(
-      <Button variant='full-width'>
+      <Button wide>
```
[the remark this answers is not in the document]
rangoo94: Yes, I was unsure as well for `wide` property.

- `fill` - goes too near `ghost` property
- `full` - may say about different behavior
- `fluid` - it's more likely what we are doing here

Then, do you agree that `fluid` will be better?

packages/phone-input/src/PhoneInput.js
rangoo94: We need to make PR to `text-mask-input` which will fix its behavior.

packages/tooltip/src/Tooltip.js
rangoo94: If you are thinking about requesting animation frame all the time, it will decrease performance a lot. Neither scrolling or resizing is a situation which is happening all the time, so why should we degrade user when he is not doing anything?

packages/calendar/src/Calendar.js
```diff
@@ -114,21 +123,26 @@ class Calendar extends React.PureComponent {
...
 
     return (
       <div className={clsName} {...passedProps}>
         <SingleDatePicker
+          hideKeyboardShortcutsPanel
           date={date}
           dayAriaLabelFormat={dayAriaLabelFormat}
           displayFormat={displayFormat}
           focused={focused}
+          firstDayOfWeek={1}
```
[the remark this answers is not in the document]
rangoo94: Could be, probably we should just make `1` as default value.

packages/checkbox/src/Checkbox.js
```diff
@@ -12,14 +12,17 @@ const propTypes = {
   /** Checkbox description */
   children: PropTypes.node,
 
+  /** Is checkbox checked? */
+  value: PropTypes.bool,
```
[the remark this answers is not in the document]
rangoo94: It is very important to keep our fields in the same format, to ensure that every field will work the same way (using `value` and `onChange`). Most importantly, we can use then wrapper like `FormField` without any problems.

packages/options-input/src/OptionsInputList.js
rangoo94: Could be, but there are problems with storybook info add-ons, when you are using external shapes.

packages/accordion/src/Accordion.js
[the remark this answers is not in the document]
michalochman: Or we can just add a note that this function is intended to work with unique lists.

packages/accordion/src/Accordion.js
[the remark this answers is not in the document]
michalochman: Fair enough

packages/button/tests/Button.test.js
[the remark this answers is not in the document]
michalochman: Yes, I think `fluid` is a good name (as in semantic-ui).

packages/checkbox/src/Checkbox.js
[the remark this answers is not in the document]
michalochman: Can you explain how it works with multiple checkboxes? For example, there's a form where you can select multiple fruits using checkboxes:
```html
<input name="fruit" value="apple">
<input name="fruit" value="banana">
<input name="fruit" value="orange">
...
```

packages/options-input/src/OptionsInputList.js
michalochman: 😕

packages/phone-input/src/PhoneInput.js
michalochman: Can we add the above as code comment? 😄

packages/tooltip/src/Tooltip.js
michalochman: It should not degrade performance at all if you return early when nothing changed – the cost of calling rAF is negligible in this scenario. I also think that using rAF to recalculate tooltip position is perfect, because it's called the same number of times as the layout is recalculated.

rangoo94: But when you will try to check if anything has changed you still have problems with recalculation. Maybe I will show you code snippet, to ensure that we are on the same page:

```js
class X extends React.Component {
  wX = window.scrollX
  wY = window.scrollY

  componentWillUnmount () {
    cancelAnimationFrame(this.tickFrame)
  }

  tick () {
    requestAnimationFrame(() => {
      const x = window.scrollX
      const y = window.scrollY

      if (x !== this.wX || y !== this.wY) {
        this.wX = x
        this.wY = y

        this.updatePosition()
      }

      tick()
    })
  }
}
```

or maybe you did want

```js
class X extends React.Component {
  componentWillUnmount () {
    cancelAnimationFrame(this.tickFrame)
  }

  _updatePosition () {
    // ...
  }

  updatePosition () {
    cancelAnimationFrame(this.updateFrame)
    this.updateFrame = requestAnimationFrame(this._updatePosition.bind(this))
}
```

?

packages/checkbox/src/Checkbox.js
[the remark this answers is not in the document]
rangoo94: It doesn't work with such yet - in case of `RadioInput`s we did make a `RadioGroup` component. In case of multiple checkboxes with values we should probably make `CheckboxGroup` as well, because it should be handled in code as an array of values

packages/button/README.md
```diff
@@ -21,15 +21,16 @@ to avoid having different versions of these in your application.
...
 Property name | Type      | Default | Description                    
 --------------|-----------|:-------:|--------------------------------
 className     | string    | n/a     | additional class name passed to wrapper
-color         | string    | n/a     | button color
-size          | string    | n/a     | button size
-variant       | string    | n/a     | button variant
+type          | string    | n/a     | button color, one of: `primary`, `secondary`, `tertiary`, `success`, `error`, `info`, `warning`, `link`
+ghost         | bool      | `false` | should this button be without solid background?
+small         | bool      | `false` | should this button be smaller than default?
+fluid          | bool      | `false` | should this button take full width?
```
maciek15: I think you just missed this additional space here :)

packages/switcher/src/Switcher.js
```diff
@@ -40,13 +13,80 @@ Switcher.propTypes = {
   /** Label for "Yes" */
   yesLabel: PropTypes.string,
```
maciek15: This probably should be `PropTypes.node` as you show in 'with icons' story

PR #122, opened 2018-06-21
[UI-68] File input

maciek15:
#### Task

I have created file input which also handles drag and drop upload. Also its file elements and header can be customized.

#### Checklist

- [x] Documentation prepared


packages/file-input/README.md
```diff
@@ -0,0 +1,48 @@
...
+buttonLabel   | string    | 'Browse Files'  | Label of browse file button.
+children      | nodes     | n/a             | Content displayed in drop box.
+className     | string    | n/a             | Additional class name.
+dropDisabled  | boolean   | `false`         | Disable drag and drop.
+filesRender   | function  | n/a             | File component renderer.
+files         | string    | n/a             | Dropped files.
+multiple      | boolean   | `true`          | Allows multiple files uploading.
+uploadLabel   | node      | n/a             | Label displayed alongside button.
+onChange      | function  | n/a             | onChange callback. Invoked when either files have been dropped or file input has changed.
+onDragEnd     | function  | n/a             | onDragEnd allback. Applies sent files and event.
```
rangoo94: `allback` ;)

packages/file-input/src/File.js
```diff
@@ -0,0 +1,58 @@
...
+const propTypes = {
+  /** File information object. */
+  file: PropTypes.object.isRequired,
+
+  /** File onRemove handler. */
+  onRemove: PropTypes.func
+}
+
+const defaultProps = {
+  onRemove: () => {}
```
rangoo94: could you instead, just not call `onRemove` if it doesn't exists?

packages/file-input/src/File.js
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Component which represents uploaded file.
+ *
+ * @param {object} props
+ * @param {object} props.file
+ * @param {function} [props.onRemove]
+ *
+ * @returns {Element|React.Element}
+ */
+const File = (props) => {
```
rangoo94: `function` for components

packages/file-input/src/FileInput.js
```diff
@@ -0,0 +1,342 @@
...
+import File from './File'
+import { getDataTransferFiles } from './utils'
+
+export const moduleName = 'file-input'
+
+/**
+ * Registers files in Map
+ *
+ * @param {array} rows
+ * @returns {Map<any, any>}
```
rangoo94: Actually, it's `Map<object, number>`

packages/file-input/src/FileInput.js
```diff
@@ -0,0 +1,342 @@
...
+
+import File from './File'
+import { getDataTransferFiles } from './utils'
+
+export const moduleName = 'file-input'
+
+/**
+ * Registers files in Map
+ *
+ * @param {array} rows
```
rangoo94: `object[]` or `Array<object>`, preferably first one

packages/file-input/src/FileInput.js
```diff
@@ -0,0 +1,342 @@
...
+ * @returns {Map<any, any>}
+ */
+export function registerElements (rows) {
```
rangoo94: Actually, it's not registering elements, it is building populated elements registry (`buildElementsRegistry`)

packages/file-input/src/FileInput.js
```diff
@@ -0,0 +1,342 @@
...
+  const newStorage = new Map()
```
rangoo94: it's worth to use `WeakMap` instead (because it will remove unnecessary values in-fly), so please use:

```js
const Constructor = typeof WeakMap === 'undefined' ? Map : WeakMap
const newStorage = new Constructor()
```

packages/file-input/src/FileInput.js
rangoo94: Actually these are not rows, they are uploaded files

packages/file-input/src/FileInput.js
```diff
@@ -0,0 +1,342 @@
...
+    this.setState({ files: newFiles, idStorage: newIdStorage, draggingOver: false })
+  }
+
+  /**
+   * Handles files removing
+   * @param {object} file
+   */
+  handleRemove (file) {
+    const { files } = this.state
+    const newFiles = files.filter(item => item !== file)
```
rangoo94: It shouldn't allow to remove files if it's controlled

packages/file-input/src/FileInput.js
```diff
@@ -0,0 +1,342 @@
...
+    const filesCls = buildClassName([moduleName, 'files-wrapper'])
+    const FileComponent = filesRender || File
+
+    return (
+      <React.Fragment>
+        <div className={filesCls}>
+          {
+            files.map((file, index) => (
+              <FileComponent
+                key={idStorage.get(file)}
```
rangoo94: It's the only place `idStorage` is used, so it could be replaced to `index` and ID Storage is not required anymore

packages/file-input/styles/config.sass
```diff
@@ -0,0 +1,30 @@
...
+$file-input-remove-color: $x-color-error !default
+
+// Text styles
+$file-input-name-font-weight: $x-weight-semibold !default
+
+// Animations
+$file-input-transition: .2s linear !default
+
+//Wrapper sizes
+$file-input-padding: 0 !default
```
rangoo94: I will fix these styles later ;)

packages/file-input/src/FileInput.js
```diff
@@ -0,0 +1,337 @@
...
+ * @property {function} [props.onDragExit]
+ * @property {function} [props.onDragLeave]
+ * @property {function} [props.onDragOver]
+ * @property {function} [props.onDragStart]
+ * @class
+ */
+class FileInput extends React.PureComponent {
+  state = {
+    draggingOver: false,
+    files: this.props.files || []
```
stuniel: You can also set up a default prop for `this.props.file`.

packages/file-input/src/FileInput.js
```diff
@@ -0,0 +1,337 @@
...
+   * Saves input reference.
+   *
+   * @param {Element|ReactElement} node
+   */
+  setInputRef = node => { this.input = node }
+
+  render () {
+    const {
+      buttonLabel, className, children, uploadLabel, dropDisabled, filesRender, multiple, onChange,
+      onDragEnd, onDragEnter, onDragExit, onDragLeave, onDragOver, onDragStart,
```
stuniel: Missing `files` prop.

PR #123, opened 2018-06-21
Feature/UI 69 countdown

PiotrGrzechnik:
#### Task

- **Issue or task referred:** UI-69

Countdown component counts down a time until a deadline. User can specify time format which will be displayed and can choose whether the '00' digits will be displayed or not.
#### Checklist

- [x] Documentation prepared


packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,165 @@
...
+import PropTypes from 'prop-types'
+
+import { buildClassName } from '@talixo/shared'
+
+const propTypes = {
+  /** Additional class name */
+  className: PropTypes.string,
+
+  /** Format of displayed date */
+  format: PropTypes.string,
```
rangoo94: After some thoughts, I think that it will be better to allow to provide a `render` function, which defaults to:

```jsx
function formatNumber (num) {
  if (isNaN(num)) {
    return '00'
  }

  if (num < 10) {
    return '0' + num
  }

  return '' + num
}

function render (props) {
  const { days, hours, minutes, seconds } = props

  return (
    <span className={prefix('countdown')}>
      <span className={prefix('countdown', 'days')}>{formatNumber(days)}</span>
      :
      <span className={prefix('countdown', 'hours')}>{formatNumber(hours)}</span>
      :
      <span className={prefix('countdown', 'minutes')}>{formatNumber(minutes)}</span>
      :
      <span className={prefix('countdown', 'seconds')}>{formatNumber(seconds)}</span>
    </span>
  )
}
```

It could also include labels by default. Then it may be much more flexible.

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,165 @@
...
+    min: 0,
+    sec: 0
+  }
+
+  /**
+   * Set interval to update counter
+   */
+  componentDidMount () {
+    const { targetDate } = this.props
+    setInterval(() => this.countTime(targetDate), 1000)
```
rangoo94: better use `setTimeout` (or maybe `requestAnimationFrame`?). When you will leave a window with `setInterval` for 10 minutes after you will come back it will immediately fire this method 600 times. Also, it will be better to run it more often than `1s`.

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /**
+   * Set each part of date in state
+   * @param {string} targetDate
+   */
+  countTime = (targetDate) => {
+    const time = Date.parse(targetDate) - Date.parse(new Date())
```
rangoo94: `Date.parse(new Date())` could be changed to just `Date.now()`

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /**
+   * Render parts of remaining time
+   * @param {number} num
+   * @param {string} time
+   *
+   * @returns {number | string}
+   */
+  renderTime = (num, time) => {
```
rangoo94: It will not be needed after `render` function prepared

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,165 @@
...
+    return num < 10 ? '0' + num : num
+  }
+
+  /**
+   * Render lables for each part of date
+   * @param {string} label
+   *
+   * @returns string
+   */
+  renderLabels = (label) => {
```
rangoo94: It will not be needed later

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,89 @@
...
+ * Component which represents Countdown.
+ *
+ * @property {object} props
+ * @property {string} [props.className]
+ * @property {string} [props.format]
+ * @property {string} props.targetDate
+ * @class {React.Element}
+ */
+class Countdown extends React.PureComponent {
+  state = {
```
michalochman: I don't think we need any other state than just `sec` here. The others can be derived from seconds.

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Set interval to update counter
+   */
+  componentDidMount () {
+    this.tick()
+  }
+
+  componentWillReceiveProps (nextProps) {
+    if (nextProps.targetDate !== this.props.targetDate) {
+      this.setState({ targetDate: nextProps.targetDate })
```
michalochman: 1) We don't need to save it to state, it's always used from props.
2) `this.tick()` should be called here instead, shouldn't it?

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,89 @@
...
+  componentWillUnmount () {
+    clearTimeout(this.tickTimeout)
+  }
+
+  tick = () => {
+    const { targetDate } = this.props
+
+    this.countTime(targetDate)
+
+    this.tickTimeout = setTimeout(this.tick, 300)
```
michalochman: `this.tickTimeout` should be cleared first, before `setTimeout` is called

packages/countdown/src/defaultFormat.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  if (num < 10) {
+    return '0' + num
+  }
+
+  return num
+}
+
+function defaultFormat (state) {
```
michalochman: The argument here should be named `props`.

rangoo94:
@PiotrGrzechnik, could you increase the amount of unit tests?

packages/countdown/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+These packages are required by `@talixo/countdown`, but you have to install them manually,
+to avoid having different versions of these in your application.
+
+## Supported props
+
+Property name | Type    | Required  | Default              | Description                    
+--------------|---------|-----------|:--------------------:|--------------------------------
+className     | string  |   no      |         n/a          | Additional class name passed to wrapper
+format        | string  |   no      | `'dd : hh : mm : ss'`| Format of displayed date
```
rangoo94: It should be updated

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,93 @@
...
+
+    clearTimeout(this.tickTimeout)
+    this.tickTimeout = setTimeout(this.tick, 300)
+  }
+
+  /**
+   * Set each part of date in state
+   * @param {string} targetDate
+   */
+  countTime = (targetDate) => {
```
rangoo94: It doesn't need to use arrow function

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,93 @@
...
+    })
+  }
+
+  render () {
+    const { time } = this.state
+    const sec = Math.floor((time / 1000) % 60)
+    const min = Math.floor((time / 1000 / 60) % 60)
+    const hours = Math.floor(time / (1000 * 60 * 60) % 24)
+    const days = Math.floor(time / (1000 * 60 * 60 * 24))
+    const finished = !(sec + min + hours + days)
```
rangoo94: isn't better to just compare `time === 0`?

packages/countdown/src/defaultFormat.js
```diff
@@ -0,0 +1,30 @@
...
+
+  return num
+}
+
+function defaultFormat (props) {
+  const { days, hours, min, sec, finished } = props
+
+  return (
+    <span className={buildClassName('countdown', null, { finished })} >
+      <span className={buildClassName('countdown', null, 'days')}>{formatNumber(days)} </span>
```
rangoo94: Wrong, `'days'`, `'hours'`, `'minutes'` and `'seconds'` are not modifiers

packages/countdown/src/defaultFormat.js
```diff
@@ -0,0 +1,30 @@
...
+       : <span className={buildClassName('countdown', null, 'hours')}>{formatNumber(hours)} </span>
```
rangoo94: Why do you need these spaces?

packages/countdown/src/defaultFormat.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  if (num < 10) {
+    return '0' + num
+  }
+
+  return num
+}
+
+function defaultFormat (props) {
```
rangoo94: The function name should tell what this function is actually doing

rangoo94: Also it could be component

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Component which represents Countdown.
+ *
+ * @property {object} props
+ * @property {string} [props.className]
+ * @property {string} [props.format]
+ * @property {string} props.targetDate
+ * @class {React.Element}
```
rangoo94: It's just a `@class`

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,93 @@
...
+const defaultProps = {
+  render: defaultFormat
+}
+
+/**
+ * Component which represents Countdown.
+ *
+ * @property {object} props
+ * @property {string} [props.className]
+ * @property {string} [props.format]
```
rangoo94: `format` is no longer used

packages/countdown/src/Countdown.js
```diff
@@ -0,0 +1,93 @@
...
+    const { time } = this.state
+    const sec = Math.floor((time / 1000) % 60)
+    const min = Math.floor((time / 1000 / 60) % 60)
+    const hours = Math.floor(time / (1000 * 60 * 60) % 24)
+    const days = Math.floor(time / (1000 * 60 * 60 * 24))
+    const finished = !(sec + min + hours + days)
+    const timeObj = {days, hours, min, sec, finished}
+
+    return (
+      <React.Fragment>
```
rangoo94: `React.Fragment` is not needed here.

packages/countdown/tests/Countdown.test.js
```diff
@@ -0,0 +1,62 @@
...
+    const wrapper = shallow(<Countdown targetDate={setDeadline(0)} />)
+    expect(wrapper.find('.talixo-countdown').at(0).hasClass('talixo-countdown--finished')).toBe(true)
+  })
+
+  it('update component when new props are differents', () => {
+    const wrapper = mount(<Countdown targetDate='2040-06-27T10:52:07.997Z' />)
+    wrapper.setProps({
+      targetDate: '2050-10-27T10:52:07.997Z'
+    })
+    expect(wrapper.prop('targetDate')).toBe('2050-10-27T10:52:07.997Z')
```
rangoo94: This unit test works for every React component.

packages/countdown/tests/Countdown.test.js
```diff
@@ -0,0 +1,62 @@
...
+      time: 'undefind'
+    })
+
+    expect(wrapper.find('Countdown').text()).toBe('00 : 00 : 00 : 00 ')
+
+    wrapper.unmount()
+  })
+
+  it('it calls clearTimeout before unmount', () => {
+    const globalTimeout = global.clearTimeout
```
rangoo94: This way, you've got not cleared timeout

packages/countdown/tests/Countdown.test.js
```diff
@@ -0,0 +1,62 @@
...
+    const spy = jest.fn()
+    global.clearTimeout = spy
+
+    const wrapper = mount(<Countdown targetDate='2040-06-27T10:52:07.997Z' />)
+
+    wrapper.instance().componentWillUnmount()
```
rangoo94: You shouldn't call these lifecycle methods, you should simply use `wrapper.unmount()`

PR #124, opened 2018-06-21
[UI-67] Fix carousel component behavior and code

rangoo94:
#### Task

- **Issue or task referred:** UI-67

#### Description

- Allow more customizable dots
- Clean up code
- Fix class names
- Fix props
- Fix styles
- Add & update stories

#### Checklist

- [x] Documentation prepared


packages/carousel/src/Carousel.js
```diff
@@ -51,17 +59,56 @@ const defaultProps = {
...
+      ? this.props.value % this.props.children.length
+      : 0,
+    transitionTime: this.props.animationTime
+  }
+
+  componentWillReceiveProps (props) {
+    // Handle change of animation time
+    if (props.animationTime !== this.props.animationTime) {
+      this.setState({
+        transitionTime: props.animationTime
```
michalochman: Why do we need that state? Can't we just read it from props?

packages/carousel/src/Carousel.js
```diff
@@ -51,17 +59,56 @@ const defaultProps = {
...
+      })
+    }
+
+    // Calculate number of slides available after changes
+    const slides = Math.ceil(props.children.length / props.perPage)
```
michalochman: If we use `ceil` here because `chunk` from lodash is using it, then it would be better to calculate [`slides` state](https://github.com/PublicInMotionGmbH/ui-kit/pull/124/files#diff-edc7c3283e1232859a182d2a6a653cccR109) first and then using `slides.length` – otherwise it may break when `chunk` implementation is changed to, e.g. `floor` or `round`

packages/carousel/src/Carousel.js
```diff
@@ -26,19 +28,25 @@ const propTypes = {
...
-  duration: PropTypes.number,
+  /** Slides animation time */
+  animationTime: PropTypes.number,
 
   /** Number of slides visible in one time */
   perPage: PropTypes.number,
 
   /** Function which render custom dots */
-  renderDots: PropTypes.func
+  renderDots: PropTypes.func,
```
michalochman: Can we also have `renderArrows`?

packages/carousel/src/Carousel.js
```diff
@@ -164,85 +234,102 @@ class Carousel extends React.PureComponent {
...
       onChange: this.handlerDot
     })
   }
 
+  renderArrows () {
+    const clsName = buildClassName([ moduleName, 'arrows' ])
+    const prevArrowClsName = buildClassName([ moduleName, 'arrow' ], null, [ 'prev' ])
+    const nextArrowClsName = buildClassName([ moduleName, 'arrow' ], null, [ 'next' ])
+
+    return (
```
michalochman: We could extract it to `Arrows` similar to `Dots`

packages/carousel/stories.js
```diff
@@ -74,83 +68,103 @@ class CustomWrapperComponent extends React.PureComponent {
...
+  >
+    <div className='tx-slide tx-slide--1'>SLIDE 1</div>
+    <div className='tx-slide tx-slide--2'>SLIDE 2</div>
+    <div className='tx-slide tx-slide--3'>SLIDE 3</div>
+  </Carousel>
+), () => ({ value: 1 }))
+
+addStory('as a gallery', readme, () => (
+  <div style={{ maxWidth: 1000, margin: '0 auto' }}>
+    <Carousel arrows dots renderDots={renderThumbnails}>
```
michalochman: Since we're not exactly rendering dots, but for example thumbnails, maybe we should rename dots to navigation?

packages/carousel/src/Carousel.js
rangoo94: When you have to finish infinite loop, there is `goImmediately` method, which is changing transition time to `0`. It could be eventually done outside of React flow

packages/carousel/src/CarouselAutoplay.js
```diff
@@ -0,0 +1,207 @@
...
+   * @param {number} props.interval
+   * @param {string} props.movement
+   * @param {number} [props.step]
+   * @param {number} [props.initialSlide]
+   */
+  constructor (props) {
+    super(props)
+
+    const carousel = React.Children.only(this.props.children)
+    const initialSlide = props.initialSlide == null ? carousel.props.value || 0 : 0
```
michalochman: I don't understand this, why is `initialSlide` set to 0, when provided via props?

PR #126, opened 2018-06-22
[UI-70] Avatar

stuniel:
#### Task

- **Issue or task referred:** UI-70

I've created an Avatar component which accepts image, icon or text as content.

#### Checklist

- [x] Documentation prepared


packages/avatar/propTypes/mutuallyExclusiveProps.js
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright (c) 2016 Airbnb
+ *
+ * This source code is licensed under the MIT license
+ */
+
+function wrapValidator (validator, typeName, typeChecker = null) {
+  return Object.assign(validator.bind(), {
+    typeName,
+    typeChecker// ,
+    // isRequired: Object.assign(validator.isRequired.bind(), {
```
rangoo94: Am I right that these comments are to remove?

packages/avatar/propTypes/mutuallyExclusiveProps.js
```diff
@@ -0,0 +1,66 @@
...
+    //   typeName,
+    //   typeChecker,
+    //   typeRequired: true
+    // })
+  })
+}
+
+// Additional required field
+function mutuallyExclusiveProps (propType, required, ...exclusiveProps) {
+  // if (typeof propType !== 'function') {
```
rangoo94: Are these comments to remove as well?

packages/avatar/propTypes/mutuallyExclusiveProps.js
```diff
@@ -0,0 +1,66 @@
...
+  // if (exclusiveProps.length < 1) {
+  //   throw new TypeError('at least one prop that is mutually exclusive with this propType is required')
+  // }
+
+  const propList = exclusiveProps.join(', or ')
+
+  const map = exclusiveProps.reduce((acc, prop) => ({ ...acc, [prop]: true }), {})
+  const countProps = (count, prop) => (count + (map[prop] ? 1 : 0))
+
+  const validator = function mutuallyExclusiveProps (props, propName, componentName, ...rest) {
```
rangoo94: Why it has the same name as it's parent?

packages/avatar/propTypes/mutuallyExclusiveProps.js
```diff
@@ -0,0 +1,66 @@
...
+      return new Error(`A ${componentName} cannot have more than one of these props: ${propList}`)
+    }
+    // Require at least one of exclusive props
+    if (required && exclusivePropCount === 0) {
+      return new Error(`A ${componentName} requires at least one of these props: ${propList}`)
+    }
+    return propType(props, propName, componentName, ...rest)
+  }
+
+  // validator.isRequired = function mutuallyExclusivePropsRequired (
```
rangoo94: Should it be removed?

packages/avatar/propTypes/mutuallyExclusiveProps.js
```diff
@@ -0,0 +1,66 @@
+/**
+ * Copyright (c) 2016 Airbnb
+ *
+ * This source code is licensed under the MIT license
```
rangoo94: I would add also URL for this code

packages/avatar/src/Avatar.js
```diff
@@ -0,0 +1,161 @@
...
+    this.buildChild = this.buildChild.bind(this)
+    this.handleImageError = this.handleImageError.bind(this)
+  }
+
+  /**
+   * Build avatar's child.
+   *
+   * @returns {React.Element}
+   */
+  buildChild () {
```
rangoo94: I am not sure, but maybe it will be work to extract this logic to separated component? Anyway, you can include `imageError` here as well:

```jsx
  buildChild () {
    const { alt, children, defaultText, defaultIcon, icon, imgProps, src, srcSet } = this.props
    const { imageError } = this.state

    if (children) {
      return children
    } else if (icon) {
      return <Icon name={icon} />
    } else if (imageError && defaultIcon) {
      return <Icon name={defaultIcon} />
    } else if (imageError && defaultText) {
      return defaultText
    } else if (imageError) {
      return '?'
    } else if (src || srcSet) {
      return (
        <img
          onError={this.handleImageError}
          src={src}
          srcSet={srcSet}
          alt={alt}
          {...imgProps}
        />
      )
    }
  }
```

packages/avatar/src/Avatar.js
```diff
@@ -0,0 +1,161 @@
...
+   *
+   * @returns {React.Element}
+   */
+  render () {
+    const { alt, children, className, defaultText, defaultIcon, icon, imgProps,
+      rounded, size, src, srcSet, style, ...passedProps } = this.props
+
+    // Build style for wrapper.
+    const wrapperStyle = {
+      height: size,
```
rangoo94: Shouldn't this `width`, `height`, `font-size` and `border-radius` be done in styles?

packages/avatar/styles/geometry.sass
```diff
@@ -0,0 +1,19 @@
+@import './config'
+
+@include module('avatar')
```
rangoo94: Use `main.sass` only for styles

packages/avatar/src/Avatar.js
stuniel: These attributes are controlled by `size` prop, that's why, in my opinion, this should be here.

packages/avatar/propTypes/mutuallyExclusiveProps.js
stuniel: I wasn't sure whether to comment out this code or to remove it. It isn't necessary in the current implementation of Avatar component, but it might be useful in the future.

packages/avatar/styles/main.sass
```diff
@@ -0,0 +1,19 @@
+@import './config'
+
+@include module('avatar')
+  position: relative
```
rangoo94:
```sass
@include module('avatar')
  position: relative
  font-size: 1.5em
  width: 1em
  height: 1em
  line-height: 1em
  text-align: center
  
  > img
    display: block
    width: 1em
    height: 1em
    object-fit: cover

  > &__inner
    line-height: 1.25em
    font-size: 0.8em
```

When image will be directly or text/icon in `&__inner`

packages/avatar/src/Avatar.js
rangoo94: @michalochman, @stuniel, what do you think?

michalochman: Yeah, looks like a render function

PR #128, opened 2018-06-22
Export sprite of country flags to external SVG file

rangoo94:
#### Task

Export sprite of country flags to external SVG file, so we do not have big 400+ kb JS file which is including flags sprite.

#### Checklist

- [x] Required unit tests prepared - *it doesn't have Provider tests, because it has a side effect (including element to Body + Ajax request), which may break different unit tests*
- [x] Documentation prepared


packages/country-flag/README.md
```diff
@@ -19,15 +19,53 @@ Your package should additionally have some extra dependencies:
...
+```jsx
+import React from 'react'
+import { CountryFlag, CountryFlagsProvider } from '@talixo/country-flag'
+
+// Using `file-loader` or `url-loader` in Webpack
+const spriteUrl = require('@talixo/country-flag/sprites/sprite.svg')
+
+function App () {
+  return (
+    <CountryFlagsProvider>
```
michalochman: `url` prop is missing

rangoo94: Good point, thanks!

rangoo94: Good point, thanks!

PR #130, opened 2018-06-28
[UI] Add possibility to make 0 value options visible in OptionsInput

maciek15:
#### Task

- **Issue or task referred:** LP-58

Added possibility to make some options visible even if their value is 0: 
- Added props handling
- Added proper styling
- Updated tests which were written incorrectly

#### Checklist

- [x] Documentation prepared

### DIFF
--- a/packages/options-input/README.md
+++ b/packages/options-input/README.md
@@ -28,14 +28,15 @@ to avoid having different versions of these in your application.
 
 It allows any property which can be passed to `div`. Additionally, it handles some differently:
 
-Property name | Type      | Default         | Description
---------------|-----------|:---------------:|--------------------------------
-className     | string    | n/a             | Additional class name passed to wrapper
-options       | Option[]  | `[]`            | Array with objects containing data passed to component
-value         | object    | self-controlled | Value of option/options
-onChange      | function  | n/a             | Event handler fired on value change
-onFocus       | function  | n/a             | Event handler fired on input open
-onBlur        | function  | n/a             | Event handler fired on input blur
+Property name     | Type      | Default         | Description
+------------------|-----------|:---------------:|--------------------------------
+className         | string    | n/a             | Additional class name passed to wrapper
+options           | Option[]  | `[]`            | Array with objects containing data passed to component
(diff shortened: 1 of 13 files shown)


packages/options-input/README.md
```diff
@@ -28,14 +28,15 @@ to avoid having different versions of these in your application.
...
-options       | Option[]  | `[]`            | Array with objects containing data passed to component
-value         | object    | self-controlled | Value of option/options
-onChange      | function  | n/a             | Event handler fired on value change
-onFocus       | function  | n/a             | Event handler fired on input open
-onBlur        | function  | n/a             | Event handler fired on input blur
+Property name     | Type      | Default         | Description
+------------------|-----------|:---------------:|--------------------------------
+className         | string    | n/a             | Additional class name passed to wrapper
+options           | Option[]  | `[]`            | Array with objects containing data passed to component
+persistentOptions | array     | n/a             | Array of options IDs which will be displayed event if their value is 0.
```
rangoo94: Actually, by default it's empty array, and it's `string[]`

packages/options-input/src/OptionsInputValue.js
```diff
@@ -30,10 +30,17 @@ const propTypes = {
     max: PropTypes.number
   })),
 
+  /** Array of options IDs which will be displayed event if their value is 0. */
+  persistentOptions: PropTypes.array,
```
rangoo94: `arrayOf(PropTypes.string)`

packages/options-input/src/OptionsInputValue.js
```diff
@@ -53,15 +60,17 @@ const DUMMY_OPTION = {
...
 
-  const elements = options.filter(x => value[x.id]).map(x => (
-    <Option
-      key={x.id}
-      option={x}
-      value={value[x.id]}
-    />
-  ))
+  const elements = options
+    .filter(x => value[x.id] || persistentOptions.indexOf(x.id) > -1)
```
rangoo94: more semantically correct is `!== -1`, because `-1` is not a correct index - it's constant, which says that it's not available

PR #131, opened 2018-06-29
[UI-72] Fix Countdown story with custom format

PiotrGrzechnik:
#### Task

- **Issue or task referred:** UI-72

Fix story in Countdown with custom format.

#### Checklist

- [x] Documentation prepared

### DIFF
--- a/packages/countdown/stories.js
+++ b/packages/countdown/stories.js
@@ -28,10 +28,8 @@ addStory('initial', readme, () => (
 addStory('with custom format', readme, () => (
   <Countdown targetDate={setDeadline(1000000000)}
     render={(props) => {
-      const arrType = ['days', 'hours', 'minutes', 'seconds']
-      return (
-        arrType.map((type, i) => `${props[type]} ${type} `)
-      )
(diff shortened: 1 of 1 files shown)


packages/countdown/stories.js
```diff
@@ -28,10 +28,9 @@ addStory('initial', readme, () => (
 addStory('with custom format', readme, () => (
   <Countdown targetDate={setDeadline(1000000000)}
     render={(props) => {
-      const arr = Object.values(props)
       const arrType = ['days', 'hours', 'minutes', 'seconds']
       return (
-        arrType.map((type, i) => `${arr[i]} ${type} `)
+        arrType.map((type, i) => `${props[type]} ${type} `)
```
rangoo94: You could use `join(' ')` directly, to show the proper behavior inside

rangoo94: Also, this way, you will not have trailing space on the end. An index is not required here as well.

```js
function renderSimpleCountdown (props) {
  const types = [ 'days', 'hours', 'minutes', 'seconds' ]

  return types.map(type => `${props[type]} ${type}`).join(' ')
}
```

PR #135, opened 2018-07-04
[UI-79] Checkbox Label should be always on right of Checkbox tick

rangoo94:
#### Task

- **Issue or task referred:** UI-79

Checkbox Label should be always on the right of Checkbox tick

#### Checklist

- [x] Documentation prepared


rangoo94:
Good question, I will check that :+1:

PR #137, opened 2018-07-04
[UI-82] Fix Calendar to have onChange feedback sent correctly

rangoo94:
#### Task

- **Issue or task referred:** UI-82

Improve Calendar update feedback

#### Checklist

- [ ] Documentation prepared

### DIFF
--- a/packages/calendar/src/Calendar.js
+++ b/packages/calendar/src/Calendar.js
@@ -130,7 +130,7 @@ class Calendar extends React.PureComponent {
     const { date } = this.state
     const nextDate = date && date.isValid() ? date.format('YYYY-MM-DD') : null
 
-    if (value != null) {
+    if (value === undefined) {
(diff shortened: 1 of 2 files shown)


packages/calendar/src/Calendar.js
```diff
@@ -118,6 +117,26 @@ class Calendar extends React.PureComponent {
...
+      this.updateTimeout = setTimeout(this.updateDateAfterBlur)
+    }
+  }
+
+  updateDateAfterBlur = () => {
+    const { value, onChange } = this.props
+    const { date } = this.state
+    const nextDate = date && date.isValid() ? date.format('YYYY-MM-DD') : null
+
+    if (value != null) {
```
maciek15: This condition causes a problem when the input is not controlled. Input value does not go back to the previous correct value because of it

PR #152, opened 2018-07-13
[UI-94] Suffix not working correctly

maciek15:
#### Task

- **Issue or task referred:** UI-94

Bug fixed: 
- TextInput position not positioned properly,
- TextInput suffix width does not calculate properly,
- OptionsInput suffixes not fully visible,
- OptionsInput RTL styles.

### DIFF
--- a/packages/number-input/styles/main.sass
+++ b/packages/number-input/styles/main.sass
@@ -3,9 +3,6 @@
 @include module('number-input')
   position: relative
 
-  &#{prefix('text-input')}--with-right #{prefix('text-input', 'input')}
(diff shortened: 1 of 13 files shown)


packages/number-input/styles/main.sass
```diff
@@ -4,7 +4,7 @@
   position: relative
 
   &#{prefix('text-input')}--with-right #{prefix('text-input', 'input')}
-    padding-right: $number-input-padding-with-stepper
+    //padding-right: $number-input-padding-with-stepper
```
rangoo94: It should be here

PR #154, opened 2018-07-19
[UI-99] Set minimum width for input in TimePicker

stuniel:
#### Task

- **Issue or task referred:** UI-99

Set minimum width for input in TimePicker


packages/time-picker/styles/config.sass
```diff
@@ -9,6 +9,7 @@ $time-picker-border-width: $text-input-border-width !default
 $time-picker-border-radius: $text-input-border-radius !default
 
 $time-picker-input-width-basis: dp(14) !default
+$time-picker-input-min-width: dp(10) !default
```
rangoo94: Could you use combine it with vertical paddings?

PR #161, opened 2018-07-30
[UI-104] Added event handler to modal backdrop.

maciek15:
#### Task

- **Issue or task referred:** UI-104

Added onClick event handler to a modal backdrop.

### DIFF
--- a/packages/modal/src/Modal.js
+++ b/packages/modal/src/Modal.js
@@ -5,12 +5,17 @@ import { buildClassName } from '@talixo/shared'
 
 import { Portal } from '@talixo/portal'
 
+export const moduleName = 'modal'
+
 const propTypes = {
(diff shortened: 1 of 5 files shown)


packages/modal/src/Modal.js
```diff
@@ -56,6 +62,13 @@ function Modal (props) {
   const modalClasses = buildClassName('modal', className, { informational }, [ type ])
   const contentClasses = buildClassName('modal-content')
 
+  const backdropClick = e => {
```
rangoo94: Actually, you can omit handler at all when there is no onClick

rangoo94: Also, handler passed to modal could be onOverlayClick as it's not about clicking modal itself

PR #164, opened 2018-08-03
[UI-98] Create color input

PiotrGrzechnik:
#### Task

- **Issue or task referred:** UI-98

I've created ColorInput component which allow to choose and show all colors in different formats: rgb, rgba, short hex, short hex with alpha, long hex, long hex with alpha, predefined. User can set alpha and manipulate HSL values. Also there is possibility to pass to component array with predefined colors and display them on a palette.

#### Checklist

- [x] Documentation prepared


rangoo94:
Also, some notes according to description:
- there isn't something like short hex with alpha; there is only #RRGGBBAA format, which is actually used mostly for IE
- does it allow also `hsl` and `hsla` values passed? (`hsl(0, 100%, 50%)`, `hsla(0, 100%, 50%, 0.5)`)

PiotrGrzechnik:
Still in progress, need more tests.

PiotrGrzechnik:
Screenshot of ColorInput with all options:
image

packages/color-input/src/ColorInput.js
```diff
@@ -0,0 +1,298 @@
...
+    this.setState({
+      color: color
+    })
+  }
+
+  /**
+   * Handle change of color
+   * @param {string} color
+   */
+  handleChangeColor (color) {
```
maciek15: There isn't any possibility to retrieve color using an external onChange function. The only way right now is to copy a string from the input.

packages/color-input/src/ColorInput.js
```diff
@@ -0,0 +1,298 @@
...
+    color: this.props.defaultColor || null,
+    error: false
+  }
+
+  /**
+   * Update state when value is provided by props.
+   *
+   * @param {object} props
+   */
+  componentWillReceiveProps (props) {
```
maciek15: There is no possibility to control the value of the color of this component using an external prop like value

packages/color-input/src/ColorInput.js
```diff
@@ -0,0 +1,298 @@
...
+    if (props.defaultColor !== this.state.color && props.defaultColor !== undefined) {
+      this.setState({ color: props.defaultColor })
```
maciek15: There is no reason to change color to the default color when the second one changes. As it says in the README description, default color should only be applied at the beginning.

packages/color-input/stories.js
```diff
@@ -0,0 +1,93 @@
...
+addStory('rgb as output format', readme, () => (
+  <ColorInput outputFormat='rgb' alpha hsl palette={palette} />
+))
+
+addStory('hex as output format', readme, () => (
+  <ColorInput outputFormat='hex' alpha hsl palette={palette} />
+))
+
+addStory('hsl as output format', readme, () => (
+  <ColorInput outputFormat='hsl' alpha hsl palette={palette} />
```
maciek15: In this story, if I open HSL input as the first one and try to change color, nothing happens

PR #166, opened 2018-08-08
[UI-108] Hide time picker by arrow

PiotrGrzechnik:
#### Task

- **Issue or task referred:** UI-108

You should put some description there about what you have done.

#### Checklist

- [x] Documentation prepared

### DIFF
--- a/packages/time-picker/src/TimeInput.js
+++ b/packages/time-picker/src/TimeInput.js
@@ -109,6 +109,37 @@ class TimeInput extends React.Component {
   }
 
   /**
+   * Handle mosue down on arrow
+   *
+   * @param {SyntheticEvent|Event} e
+   */
+  handleMouseDownArrow = (e) => {
+    const input = this.inputRef.input
+
+    input.focus()
+    e.preventDefault()
+
+    if (this.state.open) {
+      this.setState({
+        open: false
+      })
+    } else {
+      this.setState({
+        open: true
+      })
+    }
+  }
+
+  /**
+   * Set ref to TextInput
+   *
+   * @param {node} node
+   */
+  setInputRef = (node) => {
+    this.inputRef = node
+  }
+
+  /**
    * Build control arrow.
    *
    * @returns {React.Element}
@@ -120,7 +151,7 @@ class TimeInput extends React.Component {
     const arrowClsName = buildClassName([ moduleName, 'input' ], null, 'arrow')
 
     return (
-      <span className={arrowClsName}>
+      <span className={arrowClsName} onMouseDown={(e) => this.handleMouseDownArrow(e)} >
         <Icon name={open ? 'expand_less' : 'expand_more'} />
       </span>
     )
@@ -243,6 +274,7 @@ class TimeInput extends React.Component {
         value={inputValue}
         autoComplete='off'
         data-hj-whitelist={this.props['data-hj-whitelist']}
+        ref={this.setInputRef}
       />
     )
   }
--- a/packages/time-picker/src/TimeInput.js
+++ b/packages/time-picker/src/TimeInput.js
@@ -109,34 +109,16 @@ class TimeInput extends React.Component {
   }
 
   /**
-   * Handle mosue down on arrow
+   * Handle mouse down on arrow
    *
    * @param {SyntheticEvent|Event} e
    */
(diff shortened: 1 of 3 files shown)


packages/time-picker/styles/main.sass
```diff
@@ -61,6 +61,12 @@
   &__colon
     padding: $time-picker-colon-margin
 
+
+  .talixo-text-input__input:focus
```
rangoo94: Use `prefix` function

packages/time-picker/src/TimeInput.js
```diff
@@ -108,6 +108,37 @@ class TimeInput extends React.Component {
     this.setState({ inputValue, suffix })
   }
 
+  /**
+   * Handle mosue down on arrow
```
rangoo94: Typo, `mouse`

packages/time-picker/src/TimeInput.js
```diff
@@ -108,6 +108,37 @@ class TimeInput extends React.Component {
...
+   *
+   * @param {SyntheticEvent|Event} e
+   */
+  handleMouseDownArrow = (e) => {
+    const input = this.inputRef.input
```
rangoo94: I would prefer `findDOMNode(this.inputRef).querySelector('input')` to not use TextInput internal implementation.

packages/time-picker/src/TimeInput.js
```diff
@@ -108,6 +108,37 @@ class TimeInput extends React.Component {
...
+
+    input.focus()
+    e.preventDefault()
+
+    if (this.state.open) {
```
rangoo94: Whole condition is actually equivalent of:

```js
this.setState({
  open: !this.state.open
})
```

packages/time-picker/src/TimeInput.js
rangoo94: Also, this selector could eventually be on `setInputRef` level - we do not need to have an instance of `TextInput` here.

packages/time-picker/src/TimeInput.js
```diff
@@ -108,6 +108,37 @@ class TimeInput extends React.Component {
...
+  /**
+   * Handle mosue down on arrow
+   *
+   * @param {SyntheticEvent|Event} e
+   */
+  handleMouseDownArrow = (e) => {
+    const input = this.inputRef.input
+
+    input.focus()
+    e.preventDefault()
```
rangoo94: Why are you preventing the default behavior of event here? Is it required?

packages/time-picker/src/TimeInput.js
rangoo94: Also, it's not **even required** - TextInput exposes `inputRef` property, which is giving directly reference to DOM node of `input` field.

PiotrGrzechnik: In fact, now I found out, that I don't need ref to input anymore, because after click on arrow It is focused anyway.

packages/time-picker/src/TimeInput.js
PiotrGrzechnik: I need to prevent default behavior, because after click focus is lost (blur on TextInput is triggered).

packages/time-picker/src/TimeInput.js
```diff
@@ -120,7 +133,7 @@ class TimeInput extends React.Component {
     const arrowClsName = buildClassName([ moduleName, 'input' ], null, 'arrow')
 
     return (
-      <span className={arrowClsName}>
+      <span className={arrowClsName} onMouseDown={(e) => this.handleMouseDownArrow(e)} >
```
rangoo94: `onMouseDown={this.handleMouseDownArrow}` is enough

packages/time-picker/src/TimeInput.js
```diff
@@ -108,6 +108,19 @@ class TimeInput extends React.Component {
     this.setState({ inputValue, suffix })
   }
 
+  /**
+   * Handle mouse down on arrow
+   *
+   * @param {SyntheticEvent|Event} e
+   */
+  handleMouseDownArrow = (e) => {
+    e.preventDefault()
```
rangoo94: Why it does need to prevent default behavior?

PR #178, opened 2018-08-30
[UI-117] Allow null value passed in options

PiotrGrzechnik:
#### Task

- **Issue or task referred:** UI-117

Now allow null value passed in options.

#### Checklist

- [x] Documentation prepared

### DIFF
--- a/packages/radio-group/src/RadioGroup.js
+++ b/packages/radio-group/src/RadioGroup.js
@@ -93,7 +93,7 @@ class RadioGroup extends React.PureComponent {
 
     const focusableOption = selectedOption || options[0]
 
-    const optionsList = options.map(obj => (
+    const optionsList = options.map((obj, index) => (
       <RadioInput
         id={focusableOption === obj ? id : undefined}
         checked={_value === obj.value}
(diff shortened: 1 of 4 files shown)


packages/radio-group/src/RadioGroup.js
```diff
@@ -103,7 +103,7 @@ class RadioGroup extends React.PureComponent {
         name={name}
         error={error}
         onChange={this.change.bind(this, obj.value)}
-        value={obj.value}
+        value={obj.value || ''}
```
maciek15: No need to change to an empty string. Remove `isRequired` from PropTypes is enough.

packages/radio-group/tests/RadioGroup.test.js
```diff
@@ -91,4 +91,17 @@ describe('<RadioGroup />', () => {
...
+
+  it('should handle null value passed to option', () => {
+    const wrapper = mount(<RadioGroup
+      name='RadioGroup4'
+      options={[{value: null, label: 'one'}, {value: 2, label: 'two'}, {value: 3, label: 'three'}]}
+    />)
+
+    let RadioInput = wrapper.find('RadioInput')
+
+    expect(RadioInput.at(0).prop('value')).toBe('')
```
maciek15: If you apply the previous comment this on should be changed to `toBe(null)`

packages/radio-group/src/RadioGroup.js
rangoo94: It's much better to just use `index` as value

PR #179, opened 2018-08-31
[UI-127] Component for optional inputs

PiotrGrzechnik:
#### Task

- **Issue or task referred:** UI-127

Optional component which allow to user insert message in textarea. Optionally It can be collapsible with checkbox and also can be disabled or read only.

#### Checklist

- [x] Documentation prepared


packages/optional/src/Optional.js
```diff
@@ -0,0 +1,150 @@
...
+            className={clsCheckboxName}
+            onChange={(value) => this.handleCheckboxChange(value)}
+            value={visible}
+            disabled={disabled}
+          >
+            {<span className={clsLabelName}>{label}</span>}
+          </Checkbox>}
+        {!collapsible && label && <span className={clsLabelName}>{label}</span>}
+        {visible &&
+          <Textarea
```
maciek15: This should not be hardcoded as a TextArea component. Although TextArea can be a default input component, a user should have a possibility to customize this component via props.

packages/optional/src/Optional.js
```diff
@@ -0,0 +1,150 @@
...
+        className={clsName} {...passedProps}
+        name={name}
+        label={label}
+        value={value}
+        onChange={this.handleChange}
+      >
+        {collapsible &&
+          <Checkbox
+            className={clsCheckboxName}
+            onChange={(value) => this.handleCheckboxChange(value)}
```
maciek15: This can be shortened to `onChange={this.handleCheckboxChange}`. The same with an onChange attribute of TextArea.

packages/optional/src/Optional.js
```diff
@@ -0,0 +1,150 @@
...
+            value={visible}
+            disabled={disabled}
+          >
+            {<span className={clsLabelName}>{label}</span>}
```
stuniel: I don't think these curly brackets are necessary.

packages/optional/src/Optional.js
```diff
@@ -0,0 +1,150 @@
...
+        label={label}
+        value={value}
+        onChange={this.handleChange}
+      >
+        {collapsible &&
+          <Checkbox
+            className={clsCheckboxName}
+            onChange={(value) => this.handleCheckboxChange(value)}
+            value={visible}
+            disabled={disabled}
```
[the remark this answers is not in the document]
rangoo94: I think that it should be added to both, to make sure that it will be disabled correctly.

PR #188, opened 2018-09-24
[UI-97] Create PaneView component.

PiotrGrzechnik:
#### Task

- **Issue or task referred:** UI-97

PaneView component is created to display data in separete Panes. Panes can be resized vertically or horizontally. User can set default size for each Pane.
#### Checklist

- [x] Documentation prepared


packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+  componentDidMount () {
+    const { children } = this.props
+
+    if (children === undefined) return
+
+    const paneArr = []
+    let defaultSizeAll = 0
+    let numberOfNotDefaults = 0
+
+    children.map((el) => {
```
coderitual: why `map` rather than `forEach`?

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+        defaultSizeAll += el.props.defaultSize
+      } else {
+        numberOfNotDefaults += 1
+      }
+    })
+
+    const defaultSizeSingle = this.roundDecimals((100 - defaultSizeAll) / numberOfNotDefaults)
+
+    children.map((el, i) => {
+      paneArr.push({
```
coderitual: why map rather than forEach?

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+      if (el.props.defaultSize !== undefined) {
+        defaultSizeAll += el.props.defaultSize
+      } else {
+        numberOfNotDefaults += 1
+      }
+    })
+
+    const defaultSizeSingle = this.roundDecimals((100 - defaultSizeAll) / numberOfNotDefaults)
+
+    children.map((el, i) => {
```
coderitual: why map rather than forEach?

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  /**
+   * Convert to percent
+   *
+   * @param {number} size
+   * @param {number} paneViewDimension
+   * @returns {number}
+   */
+  convertToPercent (size, paneViewDimension) {
```
coderitual: we can move this function outside the component as it's helper only and doesn't need instance memers (`roundDecimals(result)` should be normal function as well)

packages/pane-view/src/Resizer.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {object} props
+ * @param {string} [props.className]
+ * @param {string} [props.split]
+ * @param {object} [props.style]
+ * @returns {React.Element}
+ */
+class Resizer extends React.PureComponent {
+  render () {
+    const { children, className, split, ...passedProps } = this.props
+
```
coderitual: unnecessary line

packages/pane-view/styles/storybook.sass
```diff
@@ -0,0 +1,19 @@
...
+@include module('pane-view')
+  border: dp(0.125) solid #000000
+  margin: dp(2.5)
+
+  & > div
+    text-align: center
+
+  &--vertical
+    height: dp(75)
+
```
coderitual: unnecessary line
```suggestion
```

packages/pane-view/tests/PaneView.test.js
```diff
@@ -0,0 +1,280 @@
...
+          <div>LEFT SIDE</div>
+        </Pane>
+        <Pane>
+          <div>RIGHT SIDE</div>
+        </Pane>
+      </PaneView>
+    )
+
+    wrapper.find('Resizer').simulate('mousedown')
+
```
coderitual: unnecessary line
```suggestion
```

packages/pane-view/tests/PaneView.test.js
```diff
@@ -0,0 +1,280 @@
...
+    expect(wrapper.state('current')).toBe(0)
+
+    document.dispatchEvent(new window.MouseEvent('mouseup'))
+
```
coderitual: unnecessary line
```suggestion
```

packages/pane-view/src/Pane.js
```diff
@@ -0,0 +1,71 @@
...
+ * @param {object} props
+ * @param {node} [props.children]
+ * @param {string} [props.className]
+ * @param {number} [props.size]
+ * @param {string} [props.split]
+ * @param {object} [props.style]
+ * @returns {React.Element}
+ */
+class Pane extends React.PureComponent {
+  buildStyle (size, split) {
```
coderitual: there is no need for this function to be part of the class (no `this` inside)

packages/pane-view/src/Pane.js
```diff
@@ -0,0 +1,71 @@
...
+    return (
+      <div className={clsName} style={wrapperStyle} {...passedProps}>
+        {children}
+      </div>
+    )
+  }
+}
+
+Pane.displayName = 'Pane'
+
```
coderitual: unnecessary line
```suggestion
```

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+              />}
+          </React.Fragment>
+        )}
+      </div>
+    )
+  }
+}
+
+PaneView.displayName = 'PaneView'
+
```
coderitual:
```suggestion
```

packages/pane-view/src/Resizer.js
```diff
@@ -0,0 +1,44 @@
...
+    const clsName = buildClassName(moduleName, className, [split])
+
+    return (
+      <span className={clsName} {...passedProps} />
+    )
+  }
+}
+
+Resizer.displayName = 'Resizer'
+
```
coderitual:
```suggestion
```

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+                split={split}
+              />}
+          </React.Fragment>
+        )}
+      </div>
+    )
+  }
+}
+
+PaneView.displayName = 'PaneView'
```
coderitual: Do we need this? I think setting this is only valuable for HOC's.
```suggestion
```

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+    this.setState({
+      paneArr
+    })
+  }
+
+  /**
+   * Handle componentWillUnmount
+   */
+  componentWillUnmount () {
+    this.unbindEventListeners()
```
coderitual: what about `handleMouseDown` listener?

packages/pane-view/src/Pane.js
```diff
@@ -0,0 +1,71 @@
...
+    return (
+      <div className={clsName} style={wrapperStyle} {...passedProps}>
+        {children}
+      </div>
+    )
+  }
+}
+
+Pane.displayName = 'Pane'
```
coderitual: Do we need this? I think setting this is only valuable for HOC's.
```suggestion
```

packages/pane-view/src/Resizer.js
```diff
@@ -0,0 +1,44 @@
...
+    const clsName = buildClassName(moduleName, className, [split])
+
+    return (
+      <span className={clsName} {...passedProps} />
+    )
+  }
+}
+
+Resizer.displayName = 'Resizer'
```
coderitual: Do we need this? I think setting this is only valuable for HOC's.
```suggestion
```

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+      paneArr: newPaneArr,
+      currentSizeVertical,
+      currentSizeHorizontal
+    })
+  }
+
+  render () {
+    const { children, className, split, style } = this.props
+    const { paneArr } = this.state
+
```
coderitual:
```suggestion
```

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+            el.size = activeOffsetWidthPercent
+          }
+        } else if (i === current + 1) {
+          el.size = this.convertToPercent(widthCombined, realPaneViewWidth)
+        }
+      })
+    }
+
+    if (split === 'vertical') {
+      newPaneArr.map((el, i) => {
```
coderitual: using `map` should lead to the new array rather than mutate elements.

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+      this.roundDecimals(nextWidth) -
+      this.roundDecimals(currentSizeHorizontal)
+    const heightCombined = this.roundDecimals(activeHeight) +
+      this.roundDecimals(nextHeight) -
+      this.roundDecimals(currentSizeVertical)
+
+    if (current === null) return
+
+    if (split === 'horizontal') {
+      newPaneArr.map((el, i) => {
```
coderitual: using map should lead to the new array rather than mutate elements.

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,309 @@
...
+    return this.roundDecimals(result)
+  }
+
+  /**
+   * Round number
+   *
+   * @param {number} num
+   * @returns {number}
+   */
+  roundDecimals (num) {
```
coderitual: this function should be outside class.

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,283 @@
...
+          <React.Fragment key={i}>
+            {React.cloneElement(child, {
+              ref: node => { this[`pane_${i}`] = node },
+              key: `pane_${i}`,
+              size: paneArr[i] && paneArr[i].size,
+              split: split
+            })}
+            {(i < children.length - 1) &&
+              <Resizer
+                ref={node => { this[`resizer_${i}`] = node }}
```
coderitual: I think it would be better to not mess around with dynamic class fields. How about using `this.resizers[i] = node`. The `length` of it might be broken but as long as you're not iterating through the list it's ok.

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,283 @@
...
+                key={`resizer_${i}`}
```
coderitual: why not just `i`? I know react  is complaining about that but it has the same meaning (little cheat :)).

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,283 @@
...
+                onMouseDown={this.handleMouseDown.bind(this, i)}
```
coderitual: it's already bound to this. The more clear intention (to pass `i` as an argument) would be using normal function instead:
`() => this.handleMouseDown(i)`

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,283 @@
...
+    document.removeEventListener('mouseup', this.handleMouseUp)
+  }
+
+  /**
+   * Handle mouse down
+   * @param {number} index
+   */
+  handleMouseDown (index) {
+    const { onDragStart } = this.props
+    const current = index
```
coderitual: I think this is unnecessary. You can just rename this variable below in `setState` or just change parameter name from `index` to `current`.

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,283 @@
...
+   * Handle componentWillUnmount
+   */
+  componentWillUnmount () {
+    this.unbindEventListeners()
+  }
+
+  /**
+   * Bind event listeners
+   */
+  bindEventListeners () {
```
coderitual: minor: I think it's more adding event listeners rather than binding (which are also present in this component)

```suggestion
  addEventListeners () {
```

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,283 @@
...
+    document.addEventListener('mousemove', this.handleMouseMove)
+    document.addEventListener('mousedown', this.handleMouseDown)
+    document.addEventListener('mouseup', this.handleMouseUp)
+  }
+
+  /**
+   * Unbind event listeners
```
coderitual: Is this our convention/lint rule to have comment above every single function? I think we don't have anything more from such description. If this is for documentation purposes that would be fine.

packages/pane-view/src/Pane.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { buildClassName } from '@talixo/shared'
+import { buildStyle } from '../utils/utils'
+
+const moduleName = 'pane'
+
+const propTypes = {
```
coderitual: `defaultSize` is missing here. Also would be great to inform user that `size` property is not allowed and it's for internal purposes only. Another idea would be to have separated component for the user and for internal usage. In such case we may avoid cloning elements inside `PaneView`.

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,283 @@
...
+    const current = index
+    this.setState({
+      current
+    })
+
+    if (onDragStart) {
+      onDragStart(current)
+    }
+
+    this.bindEventListeners()
```
coderitual: Why adding `handleMouseDown` (for the document) inside `handleMouseDown` handler (for the resizer)

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,283 @@
...
+      width: activeWidth,
+      top: activeTop,
+      height: activeHeight
+    } = activePane.getBoundingClientRect()
+    const resizersNumber = children.length - 1
+    const realPaneViewWidth = paneViewWidth - (resizersNumber * resizerWidth)
+    const realPaneViewHeight = paneViewHeight - (resizersNumber * resizerHeight)
+    const currentSizeHorizontal = e.clientX - activeLeft <= 0 ? 0 : e.clientX - activeLeft
+    const currentSizeVertical = e.clientY - activeTop <= 0 ? 0 : e.clientY - activeTop
+    const activeOffsetWidthPercent = convertToPercent(activePane.offsetWidth, realPaneViewWidth)
```
coderitual: `offsetWidth` is not equivalent for the `getBoundingClientRect`. It may cause wrong calculation (`realPanelViewWidth` is based on `gBCR`)

packages/pane-view/src/PaneView.js
```diff
@@ -0,0 +1,283 @@
...
+      roundDecimals(nextWidth) -
+      roundDecimals(currentSizeHorizontal)
+    const heightCombined = roundDecimals(activeHeight) +
+      roundDecimals(nextHeight) -
+      roundDecimals(currentSizeVertical)
+
+    if (current === null) return
+
+    if (split === 'horizontal') {
+      newPaneArr.forEach((el, i) => {
```
coderitual: if it's a new array, would be better to use `map` instead.

packages/pane-view/utils/utils.js
```diff
@@ -0,0 +1,35 @@
+/**
+   * Round number
+   *
+   * @param {number} num
+   * @returns {number}
+   */
+export function roundDecimals (num) {
+  return Math.round(num * 10000) / 10000
```
coderitual: why round? If this is for style it will be rounded anyway.

packages/pane-view/utils/utils.js
```diff
@@ -0,0 +1,35 @@
...
+  return roundDecimals(result)
+}
+
+export function buildStyle (size, split) {
+  if (size === undefined) return {}
+
+  const height = split === 'vertical' && size
+  const width = split === 'horizontal' && size
+
+  return { height: `${height}%`, width: `${width}%` }
```
coderitual: is returning `false` a documented way of not returning attribute for inline styles?

PR #192, opened 2018-10-08
[UI-136] Fix issues with children and pagination in wizard

PiotrGrzechnik:
#### Task

- **Issue or task referred:** UI-136

Fix issues with children and pagination in wizard.

#### Checklist

- [x] Documentation prepared

### DIFF
--- a/packages/wizard/src/Wizard.js
+++ b/packages/wizard/src/Wizard.js
@@ -9,7 +9,7 @@ const propTypes = {
   className: PropTypes.string,
 
   /** Children inside Wizard. Each child is the next step in Wizard */
-  children: PropTypes.node,
+  children: PropTypes.node.isRequired,
 
   /** Maximum number of displayed page buttons */
   displayedLimit: PropTypes.number,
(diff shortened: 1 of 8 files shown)


packages/wizard/src/Wizard.js
```diff
@@ -37,40 +44,38 @@ const propTypes = {
...
-        </Navigation>
-      </div>
-    )
-  }
+    render () {
+      const {children, className, displayedLimit, nextLabel, previousLabel, pagination, ...passedProps} = this.props
+      const {currentStep} = this.state
+      return (
+        <div className={buildClassName('wizard', className)} {...passedProps} >
+          {!children.length ? children : children[currentStep - 1]}
```
michalochman: Use https://reactjs.org/docs/react-api.html#reactchildrentoarray

packages/wizard/src/Wizard.js
```diff
@@ -37,40 +44,38 @@ const propTypes = {
...
+          {pagination && <Navigation type='pagination' >
```
michalochman: It would be better to pass `paginationProps` instead and pass those props to `<ControlledPagination />` below. This requires modifying pagination to allow for disabling buttons, pages and other elements.

packages/wizard/src/Wizard.js
```diff
@@ -76,5 +82,6 @@ class Wizard extends React.PureComponent {
 Wizard.displayName = 'Wizard'
```
coderitual:
```suggestion
```

packages/wizard/src/Wizard.js
```diff
@@ -76,5 +82,6 @@ class Wizard extends React.PureComponent {
...
```
coderitual:
```suggestion
```

PR #194, opened 2018-10-15
Fix problems with nesting our stylesheets namespaced in class

rangoo94:
Example:

```sass
.next-dashboard {
  @import '@talixo/text-input/styles/main.sass'
}
```


packages/typography/styles/main.sass
```diff
@@ -1,2 +1,175 @@
-@import './geometry'
-@import './theming'
+// Normalize CSS
```
coderitual: worth considering is to get rid of any generic normalization and keep everything encapsulated and independent from global styles.

PR #199, opened 2018-11-22
[UI-139] Add possibility to create horizontal bar chart

maciek15:
#### Task

- **Issue or task referred:** UI-139

Done: 
- Added possibility to create a horizontal bar chart.
- Removed obsolete Highlight component from ui-kit as right now it is available from react-vis library.
- Updated react-vis library to the newest version (1.11.5).

#### Checklist

- [x] Documentation prepared


packages/charts/tests/Chart.test.js
```diff
@@ -50,88 +50,99 @@ const defaultProps = createProps()
 const createWrapper = (props = defaultProps) => shallow(<Chart {...props} />)
```
coderitual: I know it's not a part of this PR, but I'm not sure if its the best helper function in here. What if I need to have `mount`  instead of shallow. Then I need to create separate method or use this one and add another option. Both solutions don's seem nice. Let's see what if I don't have this wrapper:

before:
`const wrapper = createWrapper()`
after:
`shallow(<Chart  {...defaultProps} />)`

before:
`const wrapper = createWrapper(props)`
after:
`shallow(<Chart  {...props} />)`

It doesn't look to bad right? In other cases it might be even better. What if I want to have only props specified and the rest from default?

`shallow(<Chart  {...defaultProps}  {...props} />)`

and If I need `mount` it as simple as

`mount(<Chart {...props} />)`

I think this wrapper adds abstraction which is not really needed in most cases.

